PR #20, opened 2023-09-27
Dashboard

atmosbear:
I think I accidentally mixed commits, but I tried to fix it 😅 Is this correct?


poundifdef:
Thanks for this! A few suggestions:

- Can you rebase this on top of `main`? Right now there is a lot of "noise" in the diff.
- I think you originally just used the CDN for Fomantic. I like that better than npm as it lets us avoid adding node as a dependency for now.

atmosbear:
OK! Will do, should be done either tonight or tomorrow :)

poundifdef:
superseded by #21

PR #24, opened 2023-10-11
Split credentials for different providers

### LINKED ISSUE #15: Separate queue vs s3 configs
Be able to use different vendors (AWS vs B2) for storage configs. Right now there is a single AWS config that is tightly coupled to both the queue and storage.

chumaumenze:
Use different AWS configs for AWS-specific services and other S3-compatible service providers. Force usage of S3 URL path style

Fixes #15


client/client.go
```diff
@@ -0,0 +1,55 @@
...
+	client.storageConfig = aws.NewConfig().
+		WithRegion(c.Storage.Region).
+		WithEndpoint(c.Storage.Endpoint).
+		WithCredentials(storageCreds).
+		WithS3ForcePathStyle(true)
+
+	return &client
+}
+
+func (c *Client) S3() *s3.S3 {
```
poundifdef, line 39: Is this thread safe?

chumaumenze: Yes, this is safe.

chumaumenze: Yes, this is not safe.

poundifdef:
Hi - I've tested this locally and everything is looking good. Thank you!

PR #31, opened 2023-10-17
Multiple CH server setup

### LINKED ISSUE #28: Multiple Clickhouse Servers
Today, there is a single Clickhouse configuration in the file but that does not allow me to add more servers in order to scale.

Therefore we want to do the following:

1. Have an interface in Go to get information about clickhouse server configurations. Right now that interface will read data from a config file, but in the future we will store information about servers in a database.

2. The configuration and interface should be able to support multiple servers, rather than just 1.

3. For now, we can just have our software choose the first Clickhouse server in a list when issuing queries. (In the future we may want to use round-robin or something else)

4. Finally, the `ClickhouseServerConfig` should be rolled into this new scheme for getting information about Clickhouse servers.

chumaumenze:
Closes #28

### DIFF
--- a/ch/server.go
+++ b/ch/server.go
@@ -1,0 +1,90 @@
+package ch
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"sync"
+	"time"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+)
+
+type (
+	ClickhouseProvider interface {
+		// FetchCredential retrieves clickhouse credential
+		FetchCredential(ctx context.Context, serverKey string) (ClickhouseCred, error)
+
+		// FetchConfig retrieve clickhouse configuration
+		FetchConfig(ctx context.Context, serverKey string) (ClickhouseConfig, error)
+	}
+
+	ClickhouseCred struct {
+		Protocol string `mapstructure:"protocol"`
+		Host     string `mapstructure:"host"`
+		HTTPPort string `mapstructure:"http_port"`
+		TCPPort  string `mapstructure:"tcp_port"`
+		Username string `mapstructure:"username"`
+		Password string `mapstructure:"password"`
+	}
+
+	ClickhouseConfig struct {
+		StoragePolicy       string `mapstructure:"storage_policy"`
+		MaxOpenConns        int    `mapstructure:"max_open_conns"`
+		MaxIdleConns        int    `mapstructure:"max_idle_conns"`
+		ConnMaxLifetimeSecs int    `mapstructure:"conn_max_lifetime"`
+	}
+
+	ClickhouseServer struct {
+		Credential ClickhouseCred   `mapstructure:"credential"`
+		Config     ClickhouseConfig `mapstructure:"config"`
+
+		clickhouse.Conn
+		mu sync.Mutex
+	}
+
+	ClickhouseServers struct {
+		Servers map[string]ClickhouseServer `mapstructure:"servers"`
+		mu      sync.Mutex
+	}
+)
+
+func (c *ClickhouseServer) Initialize(ctx context.Context) error {
+	opts := &clickhouse.Options{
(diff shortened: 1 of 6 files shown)


ch/server.go
```diff
@@ -0,0 +1,90 @@
...
+		MaxIdleConns        int    `mapstructure:"max_idle_conns"`
+		ConnMaxLifetimeSecs int    `mapstructure:"conn_max_lifetime"`
+	}
+
+	ClickhouseServer struct {
+		Credential ClickhouseCred   `mapstructure:"credential"`
+		Config     ClickhouseConfig `mapstructure:"config"`
+
+		clickhouse.Conn
+		mu sync.Mutex
```
poundifdef, line 44: Curious, why do we need this?

ch/server.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	ClickhouseServers struct {
+		Servers map[string]ClickhouseServer `mapstructure:"servers"`
+		mu      sync.Mutex
```
poundifdef, line 49: Same as above - does not look like this is used.

importer/importer.go
```diff
@@ -369,61 +366,22 @@ func (im *Importer) insertData(conn driver.Conn, bucket, key, db, table string,
...
-}
-
 func (im *Importer) consumeMessages(pid int) {
 	defer im.wg.Done()
 	defer log.Println("Stopping worker", pid)
 	log.Println("Starting worker", pid)
 
-	conn, err := im.connect()
-	if err != nil {
+	defaultServer, ok := im.Config.Clickhouse.Servers["default"]
```
poundifdef, line 374: Rather than doing this, I'd like to pass an instance of `ClickhouseProvider` into the `Importer`. Then have the default implementation fetch data from the config file.

If we do that, we could create a single `ClickhouseProvider` instance in `main.go` and pass it to any modules that need it. It would also let us avoid another refactor of the importer since we will only be accessing Clickhouse via the interface.

poundifdef (commented):
Awesome! A few comments but this is looking good.

ch/server.go
chumaumenze: Exclusion lock. This is currently not used. I will remove it if necessary.

importer/importer.go
chumaumenze: @poundifdef  this has similar effect as we are still reading from the config file. The major difference is that `*Importer.Start()` doesn't take the `ClickhouseProvider` parameter because `importer.NewImporter(&C)` already takes the config which includes the servers.

I can modify the `Start()` function to become `*Importer.Start(provider ClickhouseProvider)` and read in `Start(C.Clickhouse)`.

poundifdef: Let's do that - I'd suggest adding it as a parameter to `NewImporter()`.

chumaumenze:
@poundifdef  I refactored this. I also made a slight change to the cli flag. 

Execution would look like: `scratchdb -ingest -insert -config=config.toml`. Also, added context to properly handle signals and removed unnecessary wait groups.

main.go
```diff
@@ -16,84 +19,64 @@ import (
 )
 
 func main() {
-	log.SetFlags(log.LstdFlags | log.Lshortfile)
+	go func() {
```
poundifdef, line 22: Perhaps we can add a new flag, like `-pprof` for running this

poundifdef:
Thank you, I'll pull this down and test locally.

main.go
chumaumenze: Hehe... I forgot to clean up. Let me fix that.

chumaumenze: We can have a different main target for it, separate files and all.

poundifdef:
@chumaumenze Can you put the flag refactor as one PR, and the signal handling as another? That way we can merge smaller changes incrementally and I can review and test your original change (multiple clickhouse servers) in isolation.

PR #32, opened 2023-10-19
Use interface for getting key info

### LINKED ISSUE #27: Interface for Users and additional user data
The `[users]` section of the config file only has a key value pair (API Key -> database name). We need to be able to have more data associated with users.

1. We should have an interface for managing users (for example, `func GetUserByAPIKey() User`)
2. We should have a proper User interface. Right now that will only contain a function `GetDatabase()` but we will add more user fields in the future.
3. That interface should read data from a config file like it is today. In the future we'll want to use a database for managing user information.

poundifdef:
Closes #27


api_keys/file.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+type APIKeysFromFile struct {
+	FileName    string
+	users       map[string]string
+	lastUpdated time.Time
+	mu          sync.Mutex
+}
+
+func (k *APIKeysFromFile) readCSVToMap() (map[string]string, error) {
```
chumaumenze, line 18: This receiver can be a static function.

```suggestion
func readCSVToMap(filename string) (map[string]string, error) {
```

api_keys/file.go
```diff
@@ -0,0 +1,84 @@
...
+		user := record[1]
+		apiMap[apiKey] = user
+	}
+
+	return apiMap, nil
+}
+
+func (k *APIKeysFromFile) GetDetailsByKey(key string) (APIKeyDetails, bool) {
+	k.mu.Lock()
+	defer k.mu.Unlock()
```
chumaumenze, line 52: Are these necessary?

api_keys/file.go
```diff
@@ -0,0 +1,84 @@
...
+
+	if k.users == nil || time.Since(k.lastUpdated) > 30*time.Second {
+		users, ok := k.readCSVToMap()
```
chumaumenze, line 55: If `readCSVToMap()` is called at struct initialisation, the exclusion lock won't be necessary unless for writes.

chumaumenze (changes requested):
This looks good except for a few observations. I had no idea it'd require CSV reading.

api_keys/file.go
poundifdef: I think so. We want to be able to re-read from the user config file at runtime, and I want to ensure that there isn't a "blip" when this happens.

PR #33, opened 2023-10-23
Refactor CLI flags

chumaumenze:
Depends on #31 

Addresses https://github.com/scratchdata/ScratchDB/pull/31#issuecomment-1772906465


poundifdef:
Thank you. Please also update the README and the deploy scripts:

https://github.com/scratchdata/ScratchDB/blob/main/deploy/scratchdb.service
https://github.com/scratchdata/ScratchDB/blob/main/deploy/scratchworker.service

chumaumenze:
@poundifdef As this is a small change, we can also update the module name here? From `scratchdb` to `github.com/scratchdata/scratchdb`.

poundifdef:
> @poundifdef As this is a small change, we can also update the module name here? From `scratchdb` to `github.com/scratchdata/scratchdb`.

Good question - I would be fine with that refactor as its own PR. (The goal is to make PRs as small, independent, and self-contained as possible.)

PR #37, opened 2023-10-27
Setup Clickhouse Cluster Support

### LINKED ISSUE #29: Clickhouse Cluster Support
We want to be able to support cluster operations for creating tables.

1. When you create a database/table in Clickhouse, you have to specify an "[ON CLUSTER](https://clickhouse.com/docs/en/sql-reference/statements/create/database#on-cluster)" parameter in order to make sure the command is replicated across all database servers. You will need to run [clickhouse-keeper](https://clickhouse.com/docs/en/guides/sre/keeper/clickhouse-keeper) as part of your docker setup. keeper is Clickhouse's version of Zookeeper which allows different DB servers to coordinate. 

2. When we run a CREATE DATABASE/TABLE, or ALTER TABLE command in Clickhouse, we should be able to specify an "ON CLUSTER" parameter. We should add a new configuration param on the `User` object, `GetCluster() string` and use that as the cluster name. If that user does not have a cluster assigned (if the cluster is blank) then we should omit the ON CLUSTER query clause.

chumaumenze:
Enabled support for sharded and replicated cluster.

Resolves #29


poundifdef:
Thank you! One refactor request: please base this PR off of https://github.com/scratchdata/ScratchDB/pull/32

Then you can update the user [interface](https://github.com/scratchdata/ScratchDB/pull/32/files#diff-0c62e04a90500c8fc2bdb14fa8c9d757ecad4b684963671e92dbe7bb9c786af8R9) to have the cluster information, rather than reading it from the clickhouse config.

A physical server can have multiple clusters so by keying off the user then we can have a different setup for each person.

importer/importer.go
```diff
@@ -124,13 +124,12 @@ func (im *Importer) createCurl(sql string) string {
 	return curl
 }
 
-func (im *Importer) createDB(conn driver.Conn, db string) error {
-	sql := "CREATE DATABASE IF NOT EXISTS " + db + ";"
-	err := conn.Exec(context.Background(), sql)
-	return err
+func (im *Importer) createDB(conn driver.Conn, cluster, db string) error {
+	sql := "CREATE DATABASE IF NOT EXISTS " + db + " ON CLUSTER " + cluster + ";"
```
poundifdef, line 128: Let's do a check here for whether `cluster == ""`. If the cluster is blank then we should omit this.

Issue #11, opened 2023-08-25
Compress data as it is streamed


poundifdef:
We could compress data using a gzip writer or brotli writer (https://pkg.go.dev/github.com/andybalholm/brotli#NewWriter) as it streams in for efficiency. this could be configurable too.

Sharath-majjigi:
Hey can you provide more context to this I want to work on this.

poundifdef:
When new data is inserted, we append it to a file here: https://github.com/scratchdata/ScratchDB/blob/main/ingest/writer.go#L305

Instead of using a regular writer, we could instead append to a gzip or brotli file for compression. The requirements would be:

- Have a configuration option to choose how to compress
- Use a streaming writer
- When we consume these files, make sure we handle the compression method. 
- We want the _uncompressed_ data to have a limit (say, 1 GB) so we don't accidentally fill a disk when decompressing files

PR #38, opened 2023-10-29
Compress data #11

Sharath-majjigi:
Instead of using a regular writer, I have used  gzip or brotli file for compression.

Provided configuration option to choose how to compress
Just need a confirmation from you regarding is this the way we needed. If yes will implement the brotli compression as well.

Also need context regarding where do we decompress it while serving requests.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,7 @@ type Config struct {
 	Insert            InsertConfig                      `mapstructure:"insert"`
 	AWS               AWS                               `mapstructure:"aws"`
 	SSL               SSL                               `mapstructure:"ssl"`
+	Storage           Storage                           `mapstructure:"storage"`
 	Clickhouse        ClickhouseConfig                  `mapstructure:"clickhouse"`
 	ClickhouseServers map[string]ClickhouseServerConfig `mapstructure:"clickhouse_servers"`
 	Users             map[string]string                 `mapstructure:"users"`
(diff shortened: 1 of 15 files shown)


ingest/ingest.go
```diff
@@ -146,6 +147,11 @@ func (i *FileIngest) InsertData(c *fiber.Ctx) error {
 	if !ok {
 		writer = NewFileWriter(
 			dir,
+			i.Config.Ingest.MaxAgeSeconds,
```
poundifdef, line 150: There's no need to add these additional parameters as they are not related to file compression.

config/config.go
```diff
@@ -49,6 +49,8 @@ type IngestConfig struct {
 
 	HealthCheckPath        string `mapstructure:"health_check_path"`
 	FreeSpaceRequiredBytes int64  `mapstructure:"free_space_required_bytes"`
+	S3UploadWorkers        int    `mapstructure:"s3_upload_workers"`
+	CompressionMethod      string `mapstructure:"compression_method"`
```
poundifdef, line 53: Let's make this an enum

config/config.go
```diff
@@ -49,6 +49,8 @@ type IngestConfig struct {
 
 	HealthCheckPath        string `mapstructure:"health_check_path"`
 	FreeSpaceRequiredBytes int64  `mapstructure:"free_space_required_bytes"`
+	S3UploadWorkers        int    `mapstructure:"s3_upload_workers"`
```
poundifdef, line 52: Please make this PR independent of the one that handles `S3UploadWorkers`

ingest/ingest.go
```diff
@@ -410,6 +416,7 @@ func (i *FileIngest) Stop() error {
 	fmt.Println("Running cleanup tasks...")
 
 	// TODO: set readtimeout to something besides 0 to close keepalive connections
+	i.app.Server().ReadTimeout = 30 * time.Second
```
poundifdef, line 419: Let's think about making this change in a separate PR. The value should be configurable as well.

ingest/writer.go
```diff
@@ -301,10 +346,40 @@ func (f *FileWriter) Write(data string) error {
...
+	case "brotli":
+		// Logic for  brotli compression
+	default:
+		// No compression
+		compressedData = []byte(data + "\n")
+	}
+
 	// write data
-	_, err = f.fd.WriteString(data + "\n")
+	// _, err = f.fd.WriteString(data + "\n")
```
poundifdef, line 373: remove commented code

ingest/writer.go
```diff
@@ -301,10 +346,40 @@ func (f *FileWriter) Write(data string) error {
 		return err
 	}
 
+	var compressedData []byte
+	switch f.CompressionMethod {
+	case "gzip":
+		var buff bytes.Buffer
```
poundifdef, line 352: Can we avoid creating a new buffer and instead write directly to the file? You should be able to say `gzip.NewWriter(f.fd)` for a similar effect. Here's an example, and some discussion as well https://gist.github.com/mchirico/6147687

PR #44, opened 2023-11-06
Operate on cluster where available

chumaumenze:


importer/importer.go
```diff
@@ -137,15 +140,22 @@ func (im *Importer) createTable(server servers.ClickhouseServer, user apikeys.AP
 		storagePolicy = server.GetStoragePolicy()
 	}
 
+	engine := "MergeTree"
+	clusterStmt := ""
+	if cluster := user.GetDBCluster(); cluster != "" {
+		engine = "ReplicatedMergeTree('/clickhouse/{cluster}/tables/{database}/{table}', '{replica}')"
```
poundifdef, line 146: Where do the values for `{cluster}`, `{database}`, `{table}`, and `{replica}` come from?

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
 
       # create default database and user on starting
       CLICKHOUSE_DEFAULT_ACCESS_MANAGEMENT: 1
-      CLICKHOUSE_DB: scratchdb
+      CLICKHOUSE_DB: scratch
```
poundifdef, line 25: Should this config file also include clickhouse-keeper?

chumaumenze: @poundifdef No. Keeper is currently not enabled. I will have it in a different PR.

Let me know if you would rather have it combined here.

importer/importer.go
chumaumenze: They are taken from the macro section in Clickhouse configuration file.

docker-compose.yml
poundifdef: I think it makes sense to combine it here. Otherwise the code will not work when someone pulls it down to run locally (ON CLUSTER will fail without keeper configured.)

importer/importer.go
poundifdef: Since a clickhouse physical server will have multiple clusters and databases, should we instead do this substitution using code instead of a macro file?

chumaumenze: The path is for replication-related metadata. It is essential for coordination between shards and replicas. Keeper uses it to track all parts of the data stored in the table, the log of changes, and the states of replicas. We should have Clickhouse handle this.

What are your concerns?

importer/importer.go
```diff
@@ -137,15 +140,22 @@ func (im *Importer) createTable(server servers.ClickhouseServer, user apikeys.AP
 		storagePolicy = server.GetStoragePolicy()
 	}
 
+	engine := "MergeTree"
+	clusterStmt := ""
+	if cluster := user.GetDBCluster(); cluster != "" {
+		engine = "ReplicatedMergeTree('/clickhouse/{cluster}/tables/{shard}/{database}/{table}', '{replica}')"
```
chumaumenze, line 146: @poundifdef Following [this discussion](https://github.com/scratchdata/ScratchDB/pull/44#discussion_r1385041101), as we have  default path set in
https://github.com/scratchdata/ScratchDB/blob/dba1aea19116998b7976bbd533771c0868efd7c1/docker/clickhouse/server/config.d/cluster.xml#L2-L3

it is possible to update this to:


```suggestion
		engine = "ReplicatedMergeTree"
```

importer/importer.go
poundifdef: My understanding is that clickhouse's config has 1 value for a macro. For we could have:

```
<macros>
  <database>db1</database>
</macros>
``` 

However, what if we have 2 databases, `db1` and `db2` on a host? How would clickhouse know which value of `database` to insert?

chumaumenze: @poundifdef I was not very clear in my first answer.

The cluster and replica placeholder are gotten from the macros. The table and database placeholder are from the database and table the operation is performed on. The values are replaced with the actual database and table names where the data will be stored. 

The `{database}` is the name of the ClickHouse database that contains the table. When you create a table, you either specify the database explicitly in the `CREATE TABLE` query or use the currently selected database if you're running commands interactively). The `{database}` in the Keeper path should match the name of the database you've used to create the table.
The `{table}` is the name of the table within the ClickHouse database. This is the specific table that you are making replicated across the cluster.
The `{table}` in the Keeper path is replaced with this unique table name.

For example, say we have a cluster  called `local` and a macros (macros are defined per node/replica):

```xml
<macros>
    <cluster>local</cluster>
    <replica>1</replica>
    <shard>1</shard>
</macros>
```
and the following query:

```sql
CREATE TABLE scratch.table
ON CLUSTER '{cluster}' (
    date Date,
    key Int,
    value String
)
ENGINE = ReplicatedMergeTree('/
clickhouse/{cluster}/tables/{shard}/{database}/{table}', '{replica}')
PARTITION BY toYYYYMM(date)
ORDER BY (date, key);
```

The Keeper path for the `ReplicatedMergeTree` above would be `ReplicatedMergeTree('/
clickhouse/local/tables/1/scratch/table', '1')` for the particular replica using the macros.

Ultimately, substituting the placeholder in our code will be the same as hardcoding the keeper path which will have negative effects especially when sharding.

poundifdef: I see - I just looked it up, I didn't realize that `{table}` and `{database}` are not macros that are defined in the config. But here is the scenario:

- There is a physical clickhouse server on `1.1.1.1`
- That server has two clusters: `1` and ``

If my macro says the following: `<macros><cluster>1</cluster></macros>` then I would not be able to create a `ReplicatedMergeTree` on cluster `2`. Because the macro only has a value for `1`. In this example, we'd execute this query:

```
CREATE TABLE scratch.table
ON CLUSTER '1' (
    date Date,
    key Int,
    value String
)
ENGINE = ReplicatedMergeTree('/
clickhouse/1/tables/1/scratch/table', '1')
PARTITION BY toYYYYMM(date)
ORDER BY (date, key);
```
How would we make sure to handle this? 

The second case is if we rename a table. the clickhouse [docs](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replication) say that the built-in macros will cause problems if a table is renamed:

> The two built-in substitutions {database} and {table} can be used, they expand into the table name and the database name respectively (unless these macros are defined in the macros section). So the zookeeper path can be specified as '/clickhouse/tables/{shard}/{database}/{table}'. Be careful with table renames when using these built-in substitutions. The path in ClickHouse Keeper cannot be changed, and when the table is renamed, the macros will expand into a different path, the table will refer to a path that does not exist in ClickHouse Keeper, and will go into read-only mode.

config.toml
```diff
@@ -44,9 +44,11 @@ max_open_conns = 3
...
 conn_max_lifetime = 3600
 hosted_databases = ["scratch"]
+hosted_clusters = ["scratch"]
 
 [[users]]
 api_key = "local"
 db_user = "scratch"
 db_name = "scratch"
 db_password = "scratch"
+db_cluster = "scratch"
```
poundifdef, line 54: I finally came up with a simple solution to how to manage macros. Add 3 config settings here:

- shard_macro_name
- replica_macro_name
- (etc)

Then use these values on [this line](https://github.com/scratchdata/ScratchDB/pull/44/files#diff-93784a35b2de36099e2c63a4b0151733876c40f6ff919d2069ee91843979b846R146). So instead of 

```
ReplicatedMergeTree('/clickhouse/{cluster}/tables/{shard}/{database}/{table}', '{replica}')
```

you could say something like

```
ReplicatedMergeTree('/clickhouse/{cluster}/tables/{shard}/{database}/{table}', '{config.GetReplicaMacroName()}')
```

This would allow us to change the name of the macro per api key. A single server could have multiple independent macro values depending on the clusters that are running.

chumaumenze: I updated this.

https://github.com/scratchdata/ScratchDB/blob/c820a915179637534f123bf949d38299832acabf/importer/importer.go#L141-L164

chumaumenze: @poundifdef I updated this.

https://github.com/scratchdata/ScratchDB/blob/c820a915179637534f123bf949d38299832acabf/importer/importer.go#L141-L164

importer/importer.go
```diff
@@ -132,14 +136,28 @@ func (im *Importer) executeSQL(server servers.ClickhouseServer, sql string) erro
...
 func (im *Importer) createTable(server servers.ClickhouseServer, user apikeys.APIKeyDetails, table string) error {
+	engine := "MergeTree"
+	clusterStmt := ""
+	if cluster := user.GetDBCluster(); cluster != "" {
+		shard, replica := user.GetDBShard(), user.GetDBReplica()
+		if shard == "" || replica == "" {
+			return errors.New("distributed tables require shard and replica names")
+		}
+		engine = fmt.Sprintf(
+			"ReplicatedMergeTree('/clickhouse/%s/tables/%s/{database}/{table}', '%s')",
```
poundifdef, line 147: This should be

```suggestion
			"ReplicatedMergeTree('/clickhouse/{%s}/tables/{%s}/{database}/{table}', '{%s}')",
```

We're not defining the name of the shard/replica. We're defining the name of the _macro_ which has the shard/replica.

config.toml
```diff
@@ -43,10 +43,14 @@ password = "scratch"
...
 
 [[users]]
 api_key = "local"
-db_name = "scratchdb"
+db_name = "scratch"
 db_user = "scratch"
 db_password = "scratch"
+db_cluster = "scratch"
+db_shard = "scratch_shard"
+db_replica = "scratch_replica"
```
poundifdef, line 56: Should be: 

```suggestion
db_shard_macro = "scratch_shard"
db_replica_macro = "scratch_replica"
```

We're not specifying the shard/replica name, we're specifying the name of the macro which holds the shard/replica.

docker/clickhouse/server/config.d/macros.xml
```diff
@@ -0,0 +1,7 @@
+<clickhouse>
+    <macros>
+        <cluster>$CLUSTER</cluster>
```
poundifdef, line 3: I don't think the cluster needs to be in a macro. It can just be hard-coded in the zookeeper path.

Issue #61, opened 2023-12-12
Update configuration to be able to specify a transporter


poundifdef:
In `pkg/transport/` we have different implementations to move data from the source JSON to the destination. We want this to be configurable.

1. In `config.toml`, we should be able to support different transport implementations. Today, we just have two:
    - `Memory`
    - `QueueStorage`
2. The `QueueStorage` transport takes two additional parameters: a Queue type (SQS, Redis, RabbitMQ, etc) and a Storage type (S3, FTP, etc.)
3. Thus, the software should be able to specify different types of configurations for transport, queue, and storage types and correctly instantiate their respective objects.

Right now, the `Memory` struct does not take any parameters, and `QueueStorage` relies on two (a queue and a storage backend.) However, this is not the only way we could design it.

Another option could be to have a single implementation for transport. In the case of a memory-only transport, we would create a `MemoryQueue` and `MemoryStorage` type that is passed in.

PR #62, opened 2023-12-12
feat: add `dataTransport` config to support configuring the transporter

sd1li:
implements: #61


sd1li:
This adds a new `Config.DataTransport` field which is essentially a tagged union containing the options needed to construct the final tranporter type. I'm not sure about the name, but there already exists a `config.Transport` type so I went with the next best thing I could think of.

config.tompl
```toml
[dataTransport]
type = "memory" | "queue" | "local" | "..."
options = ...
```

`type` specifies the type of transport and is used to match `options` to its concrete type `Memory...`, `Queue...`, `Local...`, etc.

`main.go` or wherever the transport is constructed then switches on one of the concrete types and calls the appropriate constructor.

I  moved the config loading into a new function `config.Load` closer to where the `*Config` types are defined.

I also added a check for undecoded fields to prevent users from entering invalid fields resulting in the app ignoring the.

In this implementation, if `[dataTransport]` isn't specified or `[dataTransport.type` is empty, it defaults to `memory` which might not be desired.

main.go
```diff
@@ -73,16 +72,25 @@ func main() {
 	// var storageBackend filestore.StorageBackend
 
 	var dataTransport transport.DataTransport
-	// dataTransport = queuestorage.NewQueueStorageTransport(queueBackend, storageBackend)
-	// dataTransport = local..NewQueueStorageTransport(queueBackend, storageBackend)
-	dataTransport = memory.NewMemoryTransport(db)
+	switch conf.DataTransport.(type) {
```
poundifdef, line 75: Perhaps this can be broken into its own function.

config/config.go
```diff
@@ -16,6 +21,13 @@ type Config struct {
 	SQS       SQS       `toml:"sqs"`
 	API       API       `toml:"api"`
 	Transport Transport `toml:"transport"`
+
+	// DataTransportConfig specifies config for the data transporter.
+	// It will be one of the following types:
+	// - *MemoryTransportConfig
+	// - *QueueTransportConfig
+	// - *LocalTransportConfig
+	DataTransport DataTransportConfig `toml:"dataTransport"`
```
poundifdef, line 30: Let's use the existing `Transport` struct. It isn't being used right now anyway.

sd1li:
Thanks. Will take a look

config/config.go
```diff
@@ -75,10 +141,100 @@ type API struct {
...
 }
 
-type Transport struct {
-	Enabled                bool   `toml:"enabled"`
-	Workers                int    `toml:"workers"`
-	SleepSeconds           int    `toml:"sleep_seconds"`
-	DataDir                string `toml:"data"`
-	FreeSpaceRequiredBytes int64  `toml:"free_space_required_bytes"`
+// decodeTransportMemory decodes the [transport] config for `[transport.type]` = `memory`
+func decodeTransportMemory(metaData toml.MetaData, confData configData) (transport.DataTransport, error) {
```
poundifdef, line 145: This logic should live in the `main` package. Let's keep the `config` package just focused on marshaling or validating configuration files themselves.

This is what I would propose (in pseudocode):

``` go
var queue Queue
if config.SQS {
  queue = NewSQSQueue(config.SQS)
} else if config.Redis {
  queue = NewRedisQueue(config.Redis)
} else {
  queue = NewDummyQueue() // Dummy implementation which does nothing or returns an error
}

var storage Storage 
if config.S3 {
  storage = NewS3(...)
} // etc

var transport Transport
switch (config.Transport.Type) {
  case: "queuestorage":
     transport = NewQueueStorageTransport(queue, storage)
  default:
    // dummy transport which returns an error
}
```

Finally, rather than a `Memory` transport, let's instead have both a Queue and Storage type of Memory, and reuse the QueueStorage transport.

config/testdata/config/invalid/transport-invalid-queue.toml
```diff
@@ -0,0 +1,17 @@
+[transport]
```
poundifdef, line 1: If we simplify how we do configurations then we will not need extensive (or any) tests just for parsing the config file.

poundifdef (changes requested):

main.go
```diff
@@ -69,13 +71,29 @@ func main() {
 	}
 	defer db.Close()
 
-	// var queueBackend queue.QueueBackend
-	// var storageBackend filestore.StorageBackend
+	var queueBackend queue.QueueBackend
+	switch config.Transport.Queue {
+	case "memory":
+		queueBackend = memoryqueue.NewQueue()
+	case "sqs":
```
poundifdef, line 79: Remove this for now, we'll add it back in when we actually have an SQS queue.

main.go
```diff
@@ -69,13 +71,29 @@ func main() {
...
+		queueBackend = memoryqueue.NewQueue()
+	default:
+		log.Fatal().Msgf("Invalid transport.queue: %s: Expected 'memory' or 'sqs'", config.Transport.Queue)
+	}
+
+	var storageBackend filestore.StorageBackend
+	switch config.Transport.Storage {
+	case "memory":
+		storageBackend = memorystore.NewStorage()
+	case "s3":
```
poundifdef, line 90: Same as above.

main.go
```diff
@@ -69,13 +71,29 @@ func main() {
...
-	// var storageBackend filestore.StorageBackend
+	var queueBackend queue.QueueBackend
+	switch config.Transport.Queue {
+	case "memory":
+		queueBackend = memoryqueue.NewQueue()
+	case "sqs":
+		// TODO: replace with SQS queue once it's implemented
+		queueBackend = memoryqueue.NewQueue()
+	default:
+		log.Fatal().Msgf("Invalid transport.queue: %s: Expected 'memory' or 'sqs'", config.Transport.Queue)
```
poundifdef, line 83:
```suggestion
		log.Fatal().Str("queue_type", config.Transport.Queue).Msg("Invalid transport.queue")
```

main.go
```diff
@@ -69,13 +71,29 @@ func main() {
...
+
+	var storageBackend filestore.StorageBackend
+	switch config.Transport.Storage {
+	case "memory":
+		storageBackend = memorystore.NewStorage()
+	case "s3":
+		// TODO: replace with S3 storage once it's implemented
+		storageBackend = memorystore.NewStorage()
+	default:
+		log.Fatal().Msgf("Invalid transport.storage: %s: Expected 'memory' or 's3'", config.Transport.Storage)
```
poundifdef, line 94:
```suggestion
		log.Fatal().Str("storage_type", config.Transport.Storage).Msg("Invalid transport.storage")
```

main.go
```diff
@@ -69,13 +72,29 @@ func main() {
...
-	// var storageBackend filestore.StorageBackend
+	var queueBackend queue.QueueBackend
+	switch config.Transport.Queue {
+	case "memory":
+		queueBackend = memoryqueue.NewQueue()
+	case "sqs":
+		// TODO: replace with SQS queue once it's implemented
+		queueBackend = memoryqueue.NewQueue()
+	default:
+		log.Fatal().Msgf("Invalid transport.queue: %s: Expected 'memory' or 'sqs'", config.Transport.Queue)
```
sd1li, line 83: @poundifdef Instead of a dummy transports that returns an error, I opted to just fail there and then since as soon as the transport starts it's going to fail annyway.

Let me know if you'd like me to implement the dummy transport instead.

poundifdef:
We can probably remove 

```
QueueProviderName     string `toml:"queue_provider"`
StorageProviderName   string `toml:"storage_provider"`
TransportProviderName string `toml:"transport_provider"`
```

config/config.go
```diff
@@ -76,9 +78,8 @@ type API struct {
...
 
 type Transport struct {
-	Enabled                bool   `toml:"enabled"`
-	Workers                int    `toml:"workers"`
-	SleepSeconds           int    `toml:"sleep_seconds"`
-	DataDir                string `toml:"data"`
-	FreeSpaceRequiredBytes int64  `toml:"free_space_required_bytes"`
+	Queue   string `toml:"queue"`
+	Storage string `toml:"storage"`
+	DataDir string `toml:"data"`
```
sd1li, line 83: I added `DataDir` and `Workers` since they seem to be required by `queuestorage` (when ran it shows an error `mkdir : no such file or directory`).

main.go
poundifdef: Makes sense

main.go
sd1li: Done.

main.go
sd1li: Done.

main.go
sd1li: Done

main.go
sd1li: Done.

poundifdef:
Looks good to me. Thanks for all your work on this!

PR #66, opened 2023-12-14
build: setup github workflow for `go {vet,test,build}`

### LINKED ISSUE #65: Add github workflow for vet, test, build, etc.
This would help prevent new bugs from being introduced as things are refactored and new implementations are added e.g. https://github.com/scratchdata/ScratchDB/issues/63

sd1li:
Closes #65

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,29 @@
+name: Go
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: Setup
+      uses: actions/setup-go@v4
+      with:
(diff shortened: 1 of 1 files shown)


poundifdef:
What will this do exactly? I see that it runs `go vet`, etc, but will it block merges? What will it actually look like in github?

sd1li:
> What will this do exactly? I see that it runs `go vet`, etc, but will it block merges?

When a pull request is opened or new commits pushed, the jobs will run. If any of them fail, the action will be blocked.

Right now it just checks for broken builds, failing tests and vet errors but, can be configured to run any command (e.g. linters).

> What will it actually look like in github?

It will look something like this (from the Go repo):

image

image

sd1li:
Or better example at the bottom of https://github.com/superfly/flyctl/pull/3129

PR #67, opened 2023-12-14
Implement data ingest V2

### LINKED ISSUE #58: Implement data ingest in v2 branch
In the v2 refactor branch, implement the `Write()`, `StartProducer()`, and `StopProducer()`  functions in [queuestorage.go](https://github.com/scratchdata/ScratchDB/blob/v2/pkg/transport/queuestorage/queuestorage.go).

This functionality will mostly be copied from [writer.go](https://github.com/scratchdata/ScratchDB/blob/main/ingest/writer.go) in the main branch.

We do not have to use the exact code from the existing functions if there are improvements in correctness.

### `Write()`

1. Takes a `[]byte` which is JSON as input and returns `(rowID string, error)`
2. The `rowID` is automatically generated as `ulid.Make().String()`
3. This function will take the input JSON, add the fields `__row_id` and `__batch_file`, and write to a local file
4. This will also rotate the file when it gets too big. The file will be rotated when any of the following conditions is met:
    1. The file is bigger than `maxFileSize`
    2. The file is older than `maxFileAge`
    3. The file has more than `maxRows` added (this is new functionality for v2)
5. The rotation should be thread-safe 

### `StartProducer()`
1. When files are rotated, we want to do the following in the background:
    1. Upload them to S3
    2. Place a message on a queue with information about that file

### `StopProducer()`
1. When stop is called, we want to make sure that we stop performing any more writes (the Write() function should return an error)
2. We then want to make sure any remaining data is flushed to disk and then uploaded to S3 and the Queue before returning

chumaumenze:
This merge request closes #58.

- [x] Test file writer
- [x] Test queue storage
- [x] Add appropriate doc comments


pkg/transport/queuestorage/file.go
```diff
@@ -141,24 +170,76 @@ func (f *FileWriter) Write(data []byte) (n int, err error) {
 		return
 	}
 
+	f.maxRows--
 	return
 }
 
 func (f *FileWriter) WriteLn(data []byte) (n int, err error) {
 	return f.Write(append(data, '\n'))
```
sd1li, line 178: This isn't new code so just a thought... but maybe it's worth fixing this PR?

`append` either modifies the input which is unexpected or it creates a new slice thus wasting memory.

I'd do something like make the write function take a slice of things to write (`[][]byte`), then make `Write` and `WriteLn` wrap that instead.

pkg/transport/queuestorage/file.go
```diff
@@ -23,37 +45,34 @@ type FileWriter struct {
...
+func NewFileWriter(param NewFileWriterParam) (*FileWriter, error) {
+	isZero := func(x any) bool {
+		return reflect.ValueOf(x).IsZero()
+	}
 
-	MaxFileSize int64
-	MaxRows     int64
-	MaxFileAge  time.Duration
-}
+	if isZero(param.MaxFileSize) {
```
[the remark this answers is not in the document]
chumaumenze: @sd1li This pr is a wip and I should convert it to draft.

Thanks for spotting this. Initially had to do multiple comparison of different types.

pkg/transport/queuestorage/file.go
chumaumenze: This is fine because we are not modifying any existing value. If `WriteLn()` is changing any item prior to `data[:len(data)-1]` even after append, that is when corruption occurs.

Append will reslice only when there is available capacity or provision a new array with enough space.

sd1li: To clarify: if there is capacity, it will end up modifying the _underlying array_ i.e. the part after the `data` slice, unless `data`'s cap was explicitly set (which we can't guarantee).

e.g. https://go.dev/play/p/twYSbuK0v0U

pkg/transport/queuestorage/queuestorage.go
```diff
@@ -14,34 +14,36 @@ import (
...
+	fws          map[string]*FileWriter
+	fwsMu        sync.Mutex
+	closedFiles  chan FileWriterInfo
+	timeProvider func() time.Time
 
 	wg   sync.WaitGroup
 	done chan bool
 }
 
-func NewQueueStorageTransport(queue queue.QueueBackend, storage filestore.StorageBackend) *QueueStorage {
```
poundifdef, line 38: I would prefer to keep the original function signature, as both a Queue and Storage are required to create this object.

pkg/transport/queuestorage/queuestorage.go
```diff
@@ -14,34 +14,36 @@ import (
...
+type QueueStorageTransportParam struct {
+	Queue        queue.QueueBackend
+	Storage      filestore.StorageBackend
+	TimeProvider func() time.Time
```
poundifdef, line 36: What is the purpose of a `TimeProvider`? To me, it seems like the file lifespan time should just be a configuration for the QueueStorage config, ie, `expiration_seconds` or something.

pkg/transport/queuestorage/file.go
```diff
@@ -12,9 +13,30 @@ import (
...
+const (
+	MaxFileSize int64 = 100 * 1024 * 1024 // 100MB
+	MaxRows     int64 = 1_000
+	MaxFileAge        = 1 * time.Hour
+)
+
+type NewFileWriterParam struct {
+	Key    string
+	Path   string
+	Store  filestore.StorageBackend
```
poundifdef, line 25: It doesn't look like some of these parameters are used, ie, `Store` and `Queue`.

poundifdef:
Line 68 of queuestorage.go:

- Does the `delete()` function need to be wrapped in a mutex?

Line 78 of queuestorage.go:

- How many allocations are we creating compared to the original code here: https://github.com/scratchdata/ScratchDB/blob/main/ingest/writer.go#L308
- If we are unable to add a __row_id or __batch_file, we can log an error but we should not return an error

Line 189 of queuestorage.go:

- Instead of a Map, create a struct that we can pass. For example: https://github.com/scratchdata/ScratchDB/blob/main/models/models.go#L3

pkg/transport/queuestorage/file.go
```diff
@@ -23,37 +45,34 @@ type FileWriter struct {
...
+	isZero := func(x any) bool {
+		return reflect.ValueOf(x).IsZero()
+	}
 
-	MaxFileSize int64
-	MaxRows     int64
-	MaxFileAge  time.Duration
-}
+	if isZero(param.MaxFileSize) {
+		param.MaxFileSize = MaxFileSize
```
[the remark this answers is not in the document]
chumaumenze: Reasonably defaults are ideal in many cases. It would be ideal to enforce it on parameters like `key`, `path`, etc.

pkg/transport/queuestorage/testfiles/mocks/queuestorage.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"io"
+
+	"github.com/stretchr/testify/mock"
+	"scratchdata/pkg/filestore"
+	"scratchdata/pkg/queue"
+)
+
+type TestStorage struct {
```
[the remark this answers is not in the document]
chumaumenze: We can have that in-place in a different issue.

pkg/transport/queuestorage/queuestorage.go
chumaumenze: The `TimeProvider() time.Time` is essential for mocking clocks (e.g `time.Now()` et al). By doing so we can set accurately test logic that uses the time functionality.

pkg/transport/queuestorage/queuestorage.go
chumaumenze: I changed the signature as a result of https://github.com/scratchdata/ScratchDB/pull/67#discussion_r1428250058 . Although, the usage of struct param was not entirely necessary.

pkg/transport/queuestorage/file.go
```diff
@@ -175,7 +175,13 @@ func (f *FileWriter) Write(data []byte) (n int, err error) {
 }
 
 func (f *FileWriter) WriteLn(data []byte) (n int, err error) {
-	return f.Write(append(data, '\n'))
+	if n, err = f.Write(data); err != nil {
+		return
+	}
+
+	nn, err := f.Write([]byte{'\n'})
```
sd1li, line 202: There's a race condition here. If there are concurrent calls to `Write` and `WriteLn`, `\n` could end up in the wrong place.

pkg/transport/queuestorage/queuestorage.go
poundifdef: I understand. I generally prefer not to use mocks. Instead of mocking the clock, let's set up test configurations that use a 1s (or less) time interval.

pkg/transport/queuestorage/file.go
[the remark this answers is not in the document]
poundifdef: Let's think about defaults separately. For example, perhaps default values can live alongside the `config` package rather than here. If we use a more sophisticated config [library](https://github.com/spf13/viper) then we can have all our defaults in one place.

pkg/transport/queuestorage/file.go
```diff
@@ -1,164 +1,328 @@
...
+
+	return nil
 }
 
-func (f *FileWriter) Write(data []byte) (n int, err error) {
-	f.canWrite.Lock()
-	defer f.canWrite.Unlock()
+// ensureWritable checks for constraints before any write operation.
+func (f *FileWriter) ensureWritable(dataSize int64) (err error) {
+	var openNew bool
```
sd1li, line 143: It's  not clear what `openNew` signifies, so it'd help to add a comment.

But I'd remove it and just `return f.close()`. Naked returns make it hard to see what is being returned and can also lead to bugs if variables are shadowed.

pkg/transport/queuestorage/file.go
```diff
@@ -1,164 +1,328 @@
...
+}
+
+// postOps uploads the current file and queues its detail
+func (f *FileWriter) postOps() error {
+	log.Info().
+		Str("key", f.key).
+		Str("filePath", f.path).
+		Msg("uploading file")
+
+	file, err := os.Open(f.path)
```
sd1li, line 179: `file` is opened but never `.Closed()`

pkg/transport/queuestorage/file.go
```diff
@@ -1,164 +1,328 @@
...
+		return err
+	}
+
+	log.Info().
+		Str("key", f.key).
+		Str("filePath", f.path).
+		Msg("queuing file")
+
+	var bb []byte
+	if bb, err = json.Marshal(models.FileUploadMessage{
```
sd1li, line 200: I'd simplify this by moving the assignment out of the `if` statement, then there's no need for `var bb...`

pkg/transport/queuestorage/file.go
```diff
@@ -1,164 +1,328 @@
...
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if f.terminated {
+		return 0, fmt.Errorf("file writer is terminated")
+	}
+
+	rowID := ulid.Make().String()
+	if data, err = sjson.SetBytes(data, "__row_id", rowID); err != nil {
+		log.Err(err).Msg("unable to set __row_id in JSON")
```
sd1li, line 237: Should this also return an error instead of just logging it?

pkg/transport/queuestorage/file.go
```diff
@@ -1,164 +1,328 @@
...
+	}
+	if data, err = sjson.SetBytes(data, "__batch_file", f.path); err != nil {
+		log.Err(err).Msg("unable to set __batch_file in JSON")
```
sd1li, line 240: ditto

sd1li (changes requested):

sd1li:
A general thought, maybe for a future change...

Is there a reason we use `sjson.SetBytes`? AFAICT, we only use it to set to-level fields. It seems simpler to just unmarshal to `map[string]json.RawMessage`, add these fields then marshal.

pkg/transport/queuestorage/file.go
chumaumenze: We want writes to succeed regardless of both attributes. An error from this call should not hinder write.

pkg/transport/queuestorage/file.go
chumaumenze: Same as above.

pkg/transport/queuestorage/file.go
chumaumenze: Scope.

pkg/transport/queuestorage/file.go
```diff
@@ -1,138 +1,249 @@
...
 
 	return fw, nil
 }
 
-func (f *FileWriter) create() error {
+// create creates a file and all directories in its path.
+// It sets a timer which triggers close after the maxFileAge elapses.
+func (f *FileWriter) create(fileName string) error {
+	f.path = fileName
 	err := os.MkdirAll(filepath.Dir(f.path), os.ModePerm)
```
sd1li, line 102: Not new code, but this uses permission `0777` i.e. read/write/execute for every user on the system. Is this intended? it's typically `0755` or `0700` where potentially sensitive data is being stored.

sd1li (approved):
LGTM, apart from the possibly un-intended permission bits

poundifdef:
This code does not work when I run it locally:

1. I get this error:
```
10:51AM ERR queuestorage.go:125 > unable to make required directories error="mkdir : no such file or directory"
```

2. When I POST new data, nothing seems to happen - data is not uploaded, etc. 
3. When I try to stop the process (`Ctrl-C`) then the process does try to upload files, but then it hangs on the following line:

```
10:53AM INF queuestorage.go:140 > Shutting down data importer
```

What are you using as your config.toml?

poundifdef (changes requested):

chumaumenze:
@poundifdef these errors emanates from different receivers (StartConsumer & StopConsumer) which are beyond the scope of this change request.

pkg/transport/queuestorage/testdata/mocks/queuestorage.go
```diff
@@ -0,0 +1,39 @@
+package mocks
```
poundifdef, line 1: This file does not seem to be used, can we delete it?

models/models.go
```diff
@@ -31,3 +31,15 @@ type DatabaseConnection struct {
...
 }
+
+type FileUploadMessage struct {
+	// Key is the unique database connection identifier
+	Key string `json:"key"`
+
+	// Path is the S3 upload key.
+	Path string `json:"path"`
+
+	// Table is the name of the Clickhouse table
```
sd1li, line 42: nit: I'd replace `Clickhouse` with `database` and `S3` (above) with `Storage` since there are multiple implementations.

pkg/transport/queuestorage/file.go
```diff
@@ -1,138 +1,244 @@
...
 
 	return nil
 }
 
-func (f *FileWriter) countDown() {
-	defer f.wg.Done()
-	defer f.ticker.Stop()
+// ensureWritable checks for constraints before any write operation.
+func (f *FileWriter) ensureWritable(dataSize int64) error {
+	var fileInfo os.FileInfo
```
sd1li, line 142: nit: This declaration is redundant.

sd1li (approved):
LGTM, apart from the couple nits

PR #73, opened 2023-12-15
feat: implement in-memory queue and storage

### LINKED ISSUE #72: Implement in-memory queue and storage

sd1li:
Closes #72 

`StorageBackend.Download`: `io.Writer` was changed to return an `io.ReadCloser` as writer didn't look correct.


sd1li:
@poundifdef I added them to a single package so the selector is `memory.{Queue,Storage}` which reads nicer than something like `memoryqueue.Queue` or `memorystorage.Storage`

pkg/memory/queue.go
```diff
@@ -0,0 +1,48 @@
+package memory
+
+import (
+	"fmt"
+	"scratchdata/pkg/queue"
+	"sync"
+)
+
+var (
+	_ queue.QueueBackend = (*Queue)(nil)
```
poundifdef, line 10: What does this do?

sd1li: It makes compilation fail if the `queue.QueueBackend` interface changes and `Queue` doesn't implement the changes.

poundifdef: Okay. When we actually start using these objects in `main.go` this will no longer be needed. (We can delete it then.)

sd1li:
> Please use this package structure instead:
> 
>     * `pkg/filestore/memory/memory.go`
> 
>     * `pkg/queue/memory/memory.go`

Done.

poundifdef:
Thank you!

PR #80, opened 2023-12-18
build: change BackendStorage's interface to more closely match S3's

sd1li:


sd1li:
@chumaumenze maybe this is relevant too you as well?

poundifdef (approved):

PR #83, opened 2023-12-19
implement db insert for queuestorage

### LINKED ISSUE #82: Implement DB insert for QueueStorage
When new data is uploaded to our service, we want to download that data from S3 and insert into the database. This ticket is to implement the `StartConsumer()` and `StopConsumer()` functions of queuestorage.go.

- StartConsumer() will start a background job which does the following:
    - Creates a pool of workers (the number of worker should be specified in a config file)
    - Consumes messages from the queue
    - For each message, downloads the JSON file from storage
    - Calls the "WriteNDJson..." function for the database
    - After the file has been uploaded, clean up the file from local disk. If there is an error, we should log it and do not delete the file.

When we stop the go process, StopConsumer() should be called and the process should gracefully shut down. We should immediately stop consuming new message, and any data that is in progress should be uploaded.

The current logic also makes sure we have a minimum amount of space left on disk before trying to download.

There is already a basic in-memory implementation [here](https://github.com/scratchdata/ScratchDB/blob/v2/pkg/transport/memory/memory.go#L77) (though it does not ensure graceful shutdowns.)

For reference, [here](https://github.com/scratchdata/ScratchDB/blob/main/importer/importer.go#L512-L539) is the v1 logic for this process. 

In the future, we will want a better failsafe to retry uploads or requeue them if there is a failure.

sd1li:
closes #82

### DIFF
--- a/main.go
+++ b/main.go
@@ -107,14 +107,14 @@ func main() {
 			Queue:   queueBackend,
 			Storage: storageBackend,
 			WriterOpt: queuestorage.WriterOptions{
-				DataDir:     config.Transport.QueueStorage.DataDir,
+				DataDir:     config.Transport.QueueStorage.ProducerDataDir,
 				MaxFileSize: config.Transport.QueueStorage.MaxFileSizeBytes,
 				MaxRows:     config.Transport.QueueStorage.MaxRows,
 				MaxFileAge:  time.Duration(config.Transport.QueueStorage.MaxFileAgeSeconds) * time.Second,
 			},
 			DB: db,
 			// TODO: should producer and consumer use different directories?
-			DataDir:                config.Transport.QueueStorage.DataDir,
+			ConsumerDataDir:        config.Transport.QueueStorage.ConsumerDataDir,
 			DequeueTimeout:         time.Duration(config.Transport.QueueStorage.DequeueTimeoutSeconds) * time.Second,
 			FreeSpaceRequiredBytes: config.Transport.QueueStorage.FreeSpaceRequiredBytes,
 			Workers:                config.Transport.Workers,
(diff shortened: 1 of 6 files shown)


sd1li:
@poundifdef  `InsertBatchFromNDJson` takes a tableName; where does it come from? `FileUploadMessage` only contains `Key (databaseID?)` and a `Path`

poundifdef:
> @poundifdef `InsertBatchFromNDJson` takes a tableName; where does it come from? `FileUploadMessage` only contains `Key (databaseID?)` and a `Path`

The table name is supposed to come from the user's API call, however, that is not implemented yet. (Ticket to do this [here](https://github.com/orgs/scratchdata/projects/1?pane=issue&itemId=47747589)) https://github.com/scratchdata/ScratchDB/blob/v2/cmd/api/insert.go#L74-L75

The FIleUploadMessage should include a table. This is what the v1 message looks like https://github.com/scratchdata/ScratchDB/blob/main/models/models.go#L3

sd1li:
Are you adding the table field @chumaumenze ? this PR is based on https://github.com/scratchdata/ScratchDB/pull/67

poundifdef:
Does this PR need to be based on #67 ? I think it can be implemented independently, and then #67 can rebase. The only place where we should see a conflict are the FIleUploadMessage which should be easy to merge. And if you implement insert as a separate file then there shouldn't be any conflicts in queuestorage.go.

sd1li:
It's based on ``v2` because changes like`models.FileUploadMessage` which aren't yet on the `v2` branch and for easier local testing (I can `curl ...` to cal the API and eventually see the data inserted into the dummy db server), but I can try and base it on `v2` instead.

main.go
```diff
@@ -109,11 +109,28 @@ func main() {
 				MaxRows:     config.Transport.QueueStorage.MaxRows,
 				MaxFileAge:  time.Duration(config.Transport.QueueStorage.MaxFileAgeSeconds) * time.Second,
 			},
+			DB: db,
+			// TODO: should producer and consumer use different directories?
```
poundifdef, line 113: Yes

sd1li: added `produce_data_dir` and `consumer_data_dir` config options

PR #87, opened 2023-12-21
Implement endpoint for data ingest

### LINKED ISSUE #76: Implement endpoint for data ingest
The goal is to implement the logic here into v2: https://github.com/scratchdata/ScratchDB/blob/main/ingest/ingest.go#L119

This endpoint is the main API which consumes data from the user. It does the following:

1. The user must specify the `table` for data to be inserted. It can come from an `X-SCRATCHDB-TABLE` header, a `?table` query param, or the JSON body itself `{"table": "mytable"}`. Check for the table name in that order.
2. If the table is specified in a header or query param, then we assume the POST body contains our data. If the table is specified in the JSON body, we assume the user's data is contained in `{"data": {...}}`
3. Validate that the input is valid JSON
4. If the data is an array of objects, then call the Write() function for each object individually
5. Before writing JSON, we want to normalize it. This is "flattening". We have two ways of doing so, based on the `X-SCRATCHDB-FLATTEN` or `?flatten` parameter. See: https://github.com/scratchdata/ScratchDB/blob/main/ingest/ingest.go#L148
6. Once validation is done and we have determined the right configurations (table, api key, how to flatten) then we should write the json to the Writer.

chumaumenze:
closes #76 

## Tasks
- [x] Extract table name from header, query and body in respective order
- [x] Extract data from `data` field if table name is in the body
- [x] Ensure valid JSON body
- [x] Extract flattening parameter from header and query in respective order
- [x] Explode array or flatten depending on specified flatten option
- [x] Determine parameters for writing data
- [x] Testing


cmd/api/insert.go
```diff
@@ -11,86 +14,94 @@ import (
...
+	if c.Query(FlattenQuery) != "" {
+		return utils.CopyString(c.Query(FlattenQuery)), "query"
 	}
 
-	return gjson.GetBytes(c.Body(), FLATTEN_JSON).String(), "body"
+	return gjson.GetBytes(c.Body(), FlattenJson).String(), "body"
 }
 
 func (a *API) Insert(c *fiber.Ctx) error {
-	if c.QueryBool("debug", false) {
```
poundifdef, line 47: Please add this back. The purpose is to allow someone to choose to debug a single request if they want.

cmd/api/insert.go
```diff
@@ -11,86 +14,94 @@ import (
...
-			a.dataTransport.Write(connectionSetting.ID, tableName, []byte(flat))
+	parsed := gjson.ParseBytes(body)
+	if fromBody {
+		if parsed = parsed.Get("data"); !parsed.Exists() {
+			return fiber.NewError(http.StatusBadRequest, "missing required data field")
 		}
+	}
+
+	var items []gjson.Result
+	if flatAlgo == "explode" && parsed.IsArray() {
```
poundifdef, line 80: If flatAlgo == explode, we should invoke this function: https://github.com/scratchdata/ScratchDB/blob/main/ingest/flattener.go#L83

cmd/api/insert.go
```diff
@@ -11,86 +14,94 @@ import (
...
+		items = append(items, parsed.Array()...)
```
poundifdef, line 81: Does this read the entire JSON file into memory at once? If so, then we do not want to do this. Our implementation should stream JSON data to the Write() function.

cmd/api/insert.go
```diff
@@ -11,86 +14,94 @@ import (
...
+			continue
 		}
-		a.dataTransport.Write(connectionSetting.ID, tableName, []byte(flat))
+		writeErr := a.dataTransport.Write(connectionSetting.ID, tableName, []byte(flat))
+		if writeErr != nil {
+			err = errors.Join(err, flattenErr)
+		}
+	}
+	if err != nil {
+		return fiber.NewError(http.StatusMultiStatus, err.Error())
```
poundifdef, line 104: We should probably return a non-2xx error here.

cmd/api/insert_test.go
```diff
@@ -0,0 +1,136 @@
...
+			DataDir:     t.TempDir(),
+			MaxFileSize: 20,
+			MaxRows:     100,
+			MaxFileAge:  time.Hour,
+		},
+	})
+	api := NewAPIServer(cfg, db, qs)
+
+	const apiKey = "testAPIKey"
+	testCases := []struct {
```
poundifdef, line 63: These tests should really verify that we are correctly manipulating the input. For example, we should verify that our code has indeed read the table name from the correct location. Or that the output of the `flatten` algorithm is what is expected.

poundifdef (changes requested):

cmd/api/insert.go
chumaumenze: @poundifdef Which status code would you recommend here?

chumaumenze: Used `http.StatusExpectationFailed: 417`

poundifdef:
Most of this looks good. My big picture feedback is that I would like this PR to have fewer refactors which are not directly related to the feature (such as changing variables to uppercase, etc.)

cmd/api/insert.go
```diff
@@ -54,43 +45,71 @@ func (a *API) Insert(c *fiber.Ctx) error {
...
+		if tableName = gjson.GetBytes(body, TableNameJson).String(); tableName == "" {
+			return fiber.NewError(http.StatusBadRequest, "missing required table field")
+		}
+		fromBody = true
 	}
 
-	// TODO: Use actual table name from request
-	tableName := "t"
-
-	var FlattenFunc = Flatten
```
poundifdef, line 77: My intention here was to have a sort of generic `Flatten` interface, where we could have different implementations. This way the logic in the API call could be separate from how we flatten the JSON. Is there a way we can have a better separation?

cmd/api/insert.go
```diff
@@ -54,43 +45,71 @@ func (a *API) Insert(c *fiber.Ctx) error {
...
 	if connectionSetting.ID == "" {
-		return errors.New("No DB Connections set up")
+		return fiber.NewError(http.StatusUnauthorized, "no connection is set up")
 	}
-	/////
 
-	input := c.Body()
-	if !gjson.ValidBytes(input) {
-		return errors.New("Invalid JSON")
+	var (
```
poundifdef, line 62: Is there a way we can simplify the logic in this function? I realize it is complicated: we might get parameters from 3 different places (header, query param, request body) and data could be a list or an object. I would love for this function to read more linearly so we can better understand which path a piece of data is taking through the system.

sd1li: There is/was an existing `getFlattenType` method that I think did the right thing so I think it just need generalizing. Something like https://github.com/sd1li/ScratchDB/commit/fce694fee110645e273faed94fc74e019c1a34af . the names might need improving, just the gist is to have a dedicated type `DataKeys` that looks up the relevant value in the headers/query/body. I also used a n enum `DataSource` instead if "strings" for clarity.

cmd/api/insert.go
sd1li: i think a function that delegates to the appropriate implementation should be enough. Something like https://github.com/sd1li/ScratchDB/commit/a9fc7c092452684a0a75f744cae81612204da302#diff-7c8ea8145bc2dc2f03d63f005eee621fc8825e621e4e5a2eefc16c7369cda451R123

PR #91, opened 2023-12-22
add /healthcheck endpoint

### LINKED ISSUE #71: Healthcheck endpoints
For the API, we want a /healthcheck endpoint. It should behave the same way as the old endpoint: https://github.com/scratchdata/ScratchDB/blob/main/ingest/ingest.go#L62

- We should be able to specify an "unhealthy" flag for a server to manually set a server as unavailable.
- We should return an unhealthy status if the disk is full
- We should return unhealthy if there are no valid users or API keys specified

sd1li:
closes #71

### DIFF
--- a/cmd/api/healthcheck.go
+++ b/cmd/api/healthcheck.go
@@ -1,0 +1,32 @@
+package api
+
+import (
+	"os"
+	"scratchdata/util"
+
+	"github.com/gofiber/fiber/v2"
+	"github.com/rs/zerolog/log"
+)
+
+func (a *API) HealthCheck(c *fiber.Ctx) error {
+	// Check if server has been manually marked as unhealthy
+	_, err := os.Stat(a.config.HealthCheckPath)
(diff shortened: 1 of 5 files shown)


sd1li:
@poundifdef the API no longer has direct access to be api keys, users, etc. so I extended the database interface with a `HealthCheck` method.

Thinking about it more:

- should the dataTransport also be checked?
- should the user and api key _values_ be checked? currently (and on v1), only their existence is checked

poundifdef:
Good questions:

1. We do not need a queuestorage healthcheck. This is because if queuestorage is not working, we will just continue to save data to disk, so no data will be lost.
2. However, we should have monitoring to know whether queuestorage is failing. We can do this via logs and a prometheus endpoint for monitoring.
3. We do not need to check the api values themselves. We're really just checking to make sure that we don't accidentally deploy a server without a configuration file.

poundifdef (approved):

PR #92, opened 2023-12-22
start API in main.go add initial `enabled` configs

### LINKED ISSUE #89: Enable and disable services via config
The program does three things:

1. Exposes an API and runs a web server
2. Continually uploads new data (from the POST endpoint)
3. Continually inserts new data into the database

- The `config.toml` file should have an option to enable or disable each of these independently. For example, the API already has a config to do this [here](https://github.com/scratchdata/ScratchDB/blob/v2/config/config.go#L66).
- The API should have a new "read-only" mode whereby GET requests are enabled, but POST requests always return an error.
- If the API is enabled, but Uploads are disabled, then the `POST` endpoint should return an error. Additionally, the `/healthcheck` endpoint should also return an error.
- Right now, the upload service is started from `api.go` ([here](https://github.com/scratchdata/ScratchDB/blob/v2/cmd/api/api.go#L38)). Instead, we should start it from `main.go` as we do with the other daemons.

sd1li:
closes #89

### DIFF
--- a/cmd/api/api.go
+++ b/cmd/api/api.go
@@ -35,11 +35,6 @@ func (a *API) Start() error {
 		log.Fatal().Err(err).Str("path", a.config.DataDir).Msg("Unable to create data ingest directory")
 	}
 
-	err = a.dataTransport.StartProducer()
-	if err != nil {
-		return err
-	}
-
 	err = a.InitializeAPIServer()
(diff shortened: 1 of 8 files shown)


poundifdef:
Good questions. Let's put this work on pause for now until I think this through some more.

PR #94, opened 2023-12-27
make `util.ConfigToStruct` return a pointer

sd1li:
The old interface is somewhat error-prone.

Previously, given `ConfigToStruct[*T]`, `nil` would be returned if `rawConfig` was `nil`.
This resulted in `nil` being propagated through the system, eventually causing a random nil pointer crash with no obvious cause.

### DIFF
--- a/pkg/database/database.go
+++ b/pkg/database/database.go
@@ -24,7 +24,7 @@ func GetDB(config map[string]interface{}) Database {
 
 	switch configType {
 	case "static":
-		return util.ConfigToStruct[*static.StaticDB](config)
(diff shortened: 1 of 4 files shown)


sd1li:
This can be observed locally by changing `[[database_connections]].type = "clickhouse"` and removing `[database_connections.settings]` then inserting data via the API

```bash
curl -X POST 'http://localhost:3000/data?debug=1' -H 'Content-Type: application/json'  -H 'X-Api-Key: local'  -d '{"table":"t","data":{"fruit": "apple"}}'
```

Issue #90, opened 2023-12-21
Cache and clean up database connections


poundifdef:
Today, every time we perform any DB operation, we create a new connection. Instead we should cache them.

Today, we do the following to get a database connection:
 ``` go
conn := s.DB.GetDatabaseConnection(dbID)
destinations.GetDestination(conn)
```
Is there a cleaner way to do this? 

Second, once we establish a connection with a destination, we should try and maintain it, rather than closing the connection every time. Connection management should be transparent to anyone who is doing an insert/query to that destination.

sd1li:
On cleaner way:
It can be simplified with a method that calls those two functions e.g. `db.GetDestination(dbID)`  since most/all uses of `GetDestination` first calls `GetDatabaseConnection`.

On caching:
DuckDB uses `database/sql` which is a pool, and clickhouse appears to also be a pool so I think we just need to open them once and re-use the same instance everywhere.

`GetDestination` would then call a constructor e.g. `duckdb.Open` that initializes the config and opens the database then cache the result keyed by `DatabaseConnection`.

AFAICT, the `DatabaseConnection` is never changed once it's loaded from the config file, so it should  be enough to cache the destination based on the `ID`, but we can also use a deterministic/`canconical` JSON encoder to generate a stable key or just store them as a slice and do a deep equality check to find the right one.

poundifdef:
How shall we have a separation of concerns between the database (which only stores credentials) and the connection itself? Does there need to be some sort of DatabaseConnectionFactory which takes a DB as input and then maintains or caches Destination objects?

sd1li:
I don't think much needs to change in the way of architecture. We can implement it in `GetDestination`. It'd just maintain a global cache instead of creating new objects every time.

It'd look something like this:



	// pkg/.../duckdb.go
	func pkg_duckdb__Open(connectionSettings map[string]any) (*duckdb.DuckDBServer, error) {
		// ...
		s := util.ConfigToStruct[*duckdb.DuckDBServer](connectionSettings)
		connector, err := s.getConnector()
		if err != nil {
			return nil, err
		}

		s.db = sql.OpenDB(connector)
		return s, nil
	}

	var (
		destinationCache = map[string]DatabaseServer{}
	)

	func deriveDatabaseConnectionKey(dbConfig models.DatabaseConnection) string {
		return dbConfig.ID
		// or
		// JSON map order is non-deterministic so we can use a package specfically designed
		// to return deterministic output
[...]

PR #95, opened 2024-01-02
cache DatabaseServers in `destinations.GetDestination`

sd1li:
GetDestination caches the DatabaseServer object if successfully opened.

The driver.Conn and sql.DB are stored on the Clickhouse and DubDB objects respectively since they're not a single connection, but a pool.

MaxIdleConns, and other settings will likely need to be tweaked, but this prototype should be a fully working demonstration.

implements  #90

### DIFF
--- a/pkg/destinations/destinations.go
+++ b/pkg/destinations/destinations.go
@@ -1,6 +1,7 @@
 package destinations
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"scratchdata/models"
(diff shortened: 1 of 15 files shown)


sd1li:
@poundifdef It needs cleanup and maybe more testing, but this implements the idea I had in https://github.com/scratchdata/ScratchDB/issues/90

pkg/destinations/duckdb/insert.go
```diff
@@ -84,23 +84,12 @@ func (s *DuckDBServer) InsertBatchFromNDJson(table string, input io.ReadSeeker)
...
-	}
-	defer db.Close()
-
-	err = s.createTable(table, db)
-	if err != nil {
-		return err
-	}
-
-	err = s.createColumns(table, jsonTypes, db)
+	err = s.createColumns(table, jsonTypes, s.db)
```
poundifdef, line 92: In practice we probably don't need to pass `s.db` in, right?

poundifdef:
This looks good. Feel free to clean it up. 

Should we also add a `Close()` function to the destinations interface, which frees up connections and makes it invalid? When we stop the server we could call that for each connection to make sure it can safely finish its reads/writes. In the future we could have a pool of connections.

pkg/destinations/destinations.go
```diff
@@ -1,28 +1,66 @@
...
 	default:
-		return nil
+		return nil, fmt.Errorf("GetDestination: Unsupported database type: %s", dbType)
+	}
+}
+
+func GetDestination(dbConfig models.DatabaseConnection) (DatabaseServer, error) {
+	key := deriveDatabaseConnectionKey(dbConfig)
+
+	destinationCache.RLock()
```
poundifdef, line 43: would `sync.Map` be useful here?

pkg/destinations/duckdb/insert.go
sd1li: nope

pkg/destinations/destinations.go
sd1li: It's not ideal in this case because we want to block the caller, so that if N calls are made to the same ID, we only make 1 connection. Thinking about it more, I _think_ the number of connection configs will be small, so I removed the RWMutex and just use Mutex since after the first connection succeeds we'll only being doing a map lookup. If performance can become a problem it's easy to implement a slightly more complex method that blocks based on the key only.

Issue #97, opened 2024-01-05
Postgrest Syntax Support


poundifdef:
It would be great to support a subset of the [Postgrest](https://postgrest.org/en/stable/) syntax. This ticket is to implement basic query support for Clickhouse.

There is a skeleton here, which includes the start of a parser and scaffolding: https://github.com/scratchdata/ScratchDB/pull/96

The goal of this ticket is to be able to support the section "Column Filters" in this [document](https://docs.google.com/document/d/1pHwrCKpkSgO4HnNfGmUEAdHamQlhLgvIcPE4__RGFJQ/edit).

The above draft pull request already parses the query into a struct. Now the code just needs to parse that struct, create a SQL statement, and perform the query.

For this iteration, do not implement other types of filters, such as AND, OR, or limit/offset. Assume the query is just `SELECT * WHERE ...` and we will be implementing the WHERE clause for different parameters.

chumaumenze:
@sd1li, I see you opened PR #99 for this. I will reassign this to you.

PR #99, opened 2024-01-09
V2 postgrest sql

sd1li:
implements #97


models/postgrest/filteroperandtype_string.go
```diff
@@ -0,0 +1,25 @@
+// Code generated by "stringer -type=FilterOperandType"; DO NOT EDIT.
```
poundifdef, line 1: What is this for?

models/postgrest/sql.go
```diff
@@ -0,0 +1,168 @@
...
+	sqlOpertors = map[string]sqlOpertor{
+		"eq":         {S: "="},
+		"gt":         {S: ">"},
+		"gte":        {S: ">="},
+		"lt":         {S: "<"},
+		"lte":        {S: "<="},
+		"neq":        {S: "<>"},
+		"like":       {S: "like"},
+		"ilike":      {S: "ilike"},
+		"match":      {S: "~"},
```
poundifdef, line 34: I think we can just keep eq, gt, gte, lt, lte, neq, like, ilike, in, is, not, or, and, all, any. For `all` and `any`, let's translate those into the equivalent AND/OR sql clauses.

models/postgrest/filteroperandtype_string.go
sd1li: Auto-geneated `FilterOperandType.String()` method, generated by `go generate`: https://github.com/scratchdata/ScratchDB/blob/d7ff953659eba70350ab6f8fb53f08a862167161/models/postgrest/postgrest.go#L1

models/postgrest/postgrest_test.go
```diff
@@ -0,0 +1,83 @@
...
+			PopulateAST(testCase.Expression, root, parser.AST())
+
+			if testCase.Print {
+				parser.PrettyPrintSyntaxTree(testCase.Expression)
+
+				// j, err := json.MarshalIndent(root, "", "    ")
+				// if err != nil {
+				// 	t.Error(err)
+				// }
+				// log.Println(string(j))
```
chumaumenze, line 78: Is this useful?

cmd/api/postgrest.go
```diff
@@ -0,0 +1,58 @@
...
+	"scratchdata/pkg/destinations"
+
+	"github.com/gofiber/fiber/v2"
+	"github.com/gofiber/fiber/v2/utils"
+	"github.com/rs/zerolog/log"
+)
+
+func (i *API) PostgrestQuery(c *fiber.Ctx) error {
+
+	apiKey := c.Locals("apiKey").(models.APIKey)
```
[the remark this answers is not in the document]
sd1li: `apiKey` is set by `AuthMiddleware` so it should be fine here.

We could always use https://docs.gofiber.io/api/middleware/recover/ to recover from panics in general, but in this case I think a panic is desirable as it's a programmer error (the route is reachable without authentication).

chumaumenze (approved):

PR #102, opened 2024-01-12
implement AWS Redshift destination

### LINKED ISSUE #101: Redshift Destination
This ticket is to add Amazon Redshift as a destination.

Documentation for the go client library is here: https://docs.aws.amazon.com/sdk-for-go/api/service/redshift/ but it's also accessible via `database/sql` as a Postgres database

Localstack implementation docs are here: https://docs.localstack.cloud/user-guide/aws/redshift/

- Implement the DatabaseServer interface
- Inserts should happen via `COPY`

sd1li:
it's based on database/sql instead of the AWS SDK mostly because it was easier to test.
localstack doesn't appear to support Redshift or maybe only in the Pro version.

also, the AWS SDK, AFAICT, isn't designed for persistent connections so an additional connection pool would be implemented and database/sql already gives one for free.

configuration for using a local postgres instance:

```toml
[[database_connections]]
  id = "destination-1"
  account_id = "account-1"
  type = "redshift"

    [database_connections.settings]
    tls = false
    endpoint = "http://localhost:5432/scratchdb"
    username = "scratch"
    password = "..."
```

closes #101


sd1li:
This version works with real Redshift (tested with Redshift Serverless `....redshift-serverless.amazonaws.com:5439/dev`).

I don't know why the tests fail - it doesn't seem directly related to this PR, so I'll investigate some more tomorrow (they're flaky in general).

PR #108, opened 2024-01-24
Update flaky tests

sd1li:
- update actions/checkout@v4
- fix flaky sqs test
- try to fix flaky queuestorage tests
- fix failing tests after rowID logic was moved in https://github.com/scratchdata/ScratchDB/commit/52857c1bd0ffbf25aa2b3fddacb899f5b73df2f7


sd1li:
@poundifdef @chumaumenze 

In https://github.com/scratchdata/ScratchDB/pull/108/commits/b1466ca7326154a66cd0376fdaede15098aabdac I updated the FileWriter tests to remove references to `__row_id` since that log was moved in https://github.com/scratchdata/ScratchDB/commit/52857c1bd0ffbf25aa2b3fddacb899f5b73df2f7

I'm sure if that's the correct fix or not.

pkg/transport/queuestorage/file_test.go
```diff
@@ -186,14 +185,17 @@ func TestFileWriterAutoRotation(t *testing.T) {
 		param := param
 		param.MaxFileAge = 1 * time.Second
 		checkRotation(t, param, func() {
-			// +10ms to ensure rotation completes from timer expiry
-			time.Sleep(param.MaxFileAge + (10 * time.Millisecond))
+			// +1s to ensure rotation completes from timer expiry
+			time.Sleep(param.MaxFileAge + (1 * time.Second))
```
chumaumenze, line 189: Is there a reason for this increase?

chumaumenze (approved):
LGTM. Ty @sd1li.

pkg/transport/queuestorage/file_test.go
sd1li: If the machine is shared/busy, the filing writing, etc. can take longer than expected. 500ms, seemed a good value when testing with locally with `test -exec 'cpulimit ...'`, but the test already takes many seconds so I just increased it to a full second.

PR #113, opened 2024-03-13
ci: release cross platform

### LINKED ISSUE #111: Create cross-platform releases with CGO enabled
We want a github action which will cross-compile the package when a new tag is pushed and create a github release with the artifacts. 

- The package uses cgo modules, so we must be able to cross-compile for other platforms. This is probably done most easily with docker containers.
- We may use goreleaser for this. I have an example goreleaser file already but it does not work for cross-compilation. However, we are not required to use goreleaser.
- We already have the start of a github workflow, `release.yml` which uses goreleaser. You will want to modify this.

ezhil56x:
/claim #111
/fixes #111

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -10,44 +10,39 @@ builds:
     binary: scratchdata
     main: ./
     env:
+      - CGO_ENABLED=1
       - CC=x86_64-linux-gnu-gcc
       - CXX=x86_64-linux-gnu-g++
     goos:
       - linux
     goarch:
       - amd64
-    flags:
-      - -trimpath
-    ldflags:
-      - -extldflags "-lc -lrt -lpthread --static"
 
   # darwin arm64
   - id: scratchdata-darwin-arm64
     binary: scratchdata
     main: ./
     env:
+      - CGO_ENABLED=1
       - CC=oa64-clang
       - CXX=oa64-clang++
     goos:
       - darwin
     goarch:
       - arm64
-    flags:
-      - -trimpath
 
   # darwin amd64
   - id: scratchdata-darwin-amd64
     binary: scratchdata
     main: ./
     env:
+      - CGO_ENABLED=1
       - CC=oa64-clang
       - CXX=oa64-clang++
     goos:
       - darwin
     goarch:
       - amd64
-    flags:
-      - -trimpath
 
 archives:
   - id: scratchdata
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,45 +1,68 @@
-# This is an example .goreleaser.yml file with some sensible defaults.
-# Make sure to check the documentation at https://goreleaser.com
-before:
-  hooks:
-    # You may remove this if you don't use go modules.
-    - go mod tidy
-    # you may remove this if you don't need go generate
-    - go generate ./...
+project_name: scratchdata
+
+dist: .builds
+env:
+  - CGO_ENABLED=1
+
 builds:
-  - env:
-      - CGO_ENABLED=1
+  # linux amd64
+  - id: scratchdata-linux-amd64
+    binary: scratchdata
+    main: ./
+    env:
+      - CC=x86_64-linux-gnu-gcc
+      - CXX=x86_64-linux-gnu-g++
     goos:
       - linux
-      - windows
+    goarch:
+      - amd64
+    flags:
+      - -trimpath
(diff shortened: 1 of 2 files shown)


ezhil56x:
@poundifdef
Can you review once?

poundifdef:
Thank you. Can you trigger the run on your fork so I can verify that the workflow works without errors? https://github.com/ezhil56x/scratchdata/actions/workflows/release.yml

ezhil56x:
@poundifdef Is the `.goreleaser.yaml` correct or need to fix that too?

poundifdef:
goreleaser.yaml will need to be fixed. local cross-compile is failing because the package depends on cgo. So you'll need to find a different way to compile. You will probbly need platform-specific docker files. see https://goreleaser.com/limitations/cgo/

ezhil56x:
@poundifdef 

Can you provide a list of os with architecture that should be supported by GitHub actions for building?

poundifdef:
We only need Linux amd64 and Mac (arm, amd64.)

Ideally we’d have windows as well but if that is more of a pain then we can skip for now.

ezhil56x:
Thanks, will get it ready by EOD

ezhil56x:
@poundifdef Can you provide the steps to manually build for Linux amd64 and Mac (arm, amd64.) from source?

poundifdef:
If you pull down the code locally, you should be able to just run “go build .”

ezhil56x:
@poundifdef Finally done! Can you review once?

Here is the [Workflow Run](https://github.com/ezhil56x/scratchdata/actions/runs/8286241145/job/22675778776) and here is the [Release](https://github.com/ezhil56x/scratchdata/releases/tag/v2.8) triggered by the GitHub workflow

poundifdef:
Thank you. I believe we are almost there. The code depends on some dynamically-loaded libraries which are not present in the current linux build.

On linux, if I run this with your build, here is the result:

```
# ldd ./scratchdata
        not a dynamic executable
```

However, this is the expected result:

```
# ldd ./scratchdata
        linux-vdso.so.1 (0x00007ffea43dc000)
        libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fafc8c00000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fafc8eff000)
        libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007fafc8edf000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fafc8a1f000)
        /lib64/ld-linux-x86-64.so.2 (0x00007fafc8fe4000)
```

If you run the code from the release, and then run this, then the code will throw an error: `curl 'localhost:8080/api/data/query?query=select%201&api_key=local'`

.goreleaser.yaml
```diff
@@ -1,45 +1,68 @@
...
+      - CXX=x86_64-linux-gnu-g++
     goos:
       - linux
-      - windows
+    goarch:
+      - amd64
+    flags:
+      - -trimpath
+    ldflags:
+      - -extldflags "-lc -lrt -lpthread --static"
```
poundifdef, line 22: If we compile with these flags, then I get the same error:

```
CGO_ENABLED=1 go build --ldflags="-extldflags=-static" .
```

So I think the `--static` is the main problem here. Do we need any of these flags or ldflags to be set?

ezhil56x: @poundifdef Not sure, do the project requires any of those. If not we can remove them

ezhil56x:
@poundifdef 

Are `darwin` builds for `arm64` and `amd64` working as expected?
Did you get any issue when running with trimpath? If flags and ldflags are not required we can them I guess

ezhil56x:
Seems to be working after removing `ldflags` and `flags` and with `CGO_ENABLE=1`

Here is [Release](https://github.com/ezhil56x/scratchdata/releases/tag/v2.8.2) without `flags` and `ldflags` and with `CGO_ENABLED=1`


https://github.com/scratchdata/scratchdata/assets/103899034/5d7d18ec-6bf7-4b2a-9d50-3e427c873081

poundifdef:
This looks good. Thank you for all of the work you put into making these releases work as expected. Approved.

ezhil56x:
Thanks for assigning ✨

PR #114, opened 2024-03-14
Update readme.md

sujeetpillai:
Slight change in command. The api_key should follow a '?' and not an '&'


poundifdef:
Good catch, thank you!

PR #119, opened 2024-03-16
Redshift Destination

### LINKED ISSUE #117: Redshift Destination
We want to be able to read and write data from RedShift in bulk. Here are the steps to accomplish that:

## Step 1: Create new RedShift Destination interface

- [ ] Add a new package in [pkg/destinations](https://github.com/scratchdata/scratchdata/tree/main/pkg/destinations) called `redshift`
- [ ] Update [destinations.go](https://github.com/scratchdata/scratchdata/blob/main/pkg/destinations/destinations.go#L69-L74) to be able to connect to the new redshift destination.
- [ ] Implement the functions in the [Destination interface](https://github.com/scratchdata/scratchdata/blob/main/pkg/destinations/destinations.go#L20) for the new `redshift` package. For this first step, you may stub these functions (`panic("not implemented")`)

## Step 2: Implement Queries from RedShift

Set up configuration parameters for RedShift.

``` yaml
databases:
  - type: redshift
    api_keys:
      - test_api_key
    settings:
      delete_from_s3: true

      # Redshift connection
      redshift_host: myredshiftcluster.amazonaws.com 
      redshift_port: 5439 
      redshift_user: myuser 
      redshift_password: mypassword 
      redshift_dbname: mydb

      # S3 bucket
      s3_region: "us-east-1"
      s3_access_key_id: "ACCESS_KEY_ID"
      s3_secret_access_key: "SECRET_ACCESS_KEY"
      s3_bucket: "bucket_name"
```

- [ ] The `Close()` function should clean up the RedShift connection, and we should also close any open handlers related to executing the query.

- [ ] Implement the `CreateColumns()` function. This should read the input file and alter the RedShift table to add columns if they do not exist. [Here](https://github.com/scratchdata/scratchdata/blob/main/pkg/destinations/duckdb/insert.go#L79) is a reference for how to do this.

## Implement Data Insertion

We want to bulk upload data to RedShift from the input file. This means implementing the `InsertFromNDJsonFile()` function. We will do this by uploading data to S3, loading from S3 to RedShift, and then optinally deleting the file from S3

- [ ] Data should be streamed to S3 in the location provided by s3. Do not load the data into memory when uploading .
- [ ] Insert data with the [COPY](https://docs.aws.amazon.com/redshift/latest/dg/r_COPY.html) sql command. The input type is newline-delimited JSON.
- [ ] If the `delete_from_s3` flag is true, then delete the file from S3 if the copy command has been successful.

Rutik7066:
closes #117 
/claim #117


scratchdata
poundifdef, line 1: This should not be committed

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,133 @@
...
+		s3Config.s3_access_key_id,
+		s3Config.s3_secret_access_key,
+		s3Config.s3_region,
+	)
+	_, err = s.conn.Exec(copyCommand)
+	if err != nil {
+		return err
+	}
+
+	// if configOptions.Database.Settings.delete_from_s3 == true {
```
poundifdef, line 126: Should this be uncommented?

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,133 @@
...
+	// 	err = storage.Delete(fileName)
+	// 	if err != nil {
+	// 		return err
```
poundifdef, line 129: If there is an error here, let's log an error but still return nil. This is because the data has been successfully inserted, the only thing that has failed is cleanup.

pkg/destinations/redshift/query.go
```diff
@@ -0,0 +1,108 @@
...
+			val := values[i]
+			b, ok := val.([]byte)
+			if ok {
+				v = string(b)
+			} else {
+				v = val
+			}
+			row[col] = v
+		}
+		result = append(result, row)
```
poundifdef, line 48: This is appending the entire result set to a slice in memory. We do not want to do that since the result set could be very large. Instead, we want to stream data as the results are being generated.

From the issue #117:

> Data should be streamed to the writer and avoid allocations as much as possible. Do not buffer all data into memory before returning to the client. Do not generate the entire JSON/CSV payload in memory.

pkg/destinations/redshift/redshift.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"database/sql"
+	"fmt"
+
+	_ "github.com/lib/pq"
+	"github.com/scratchdata/scratchdata/util"
+)
+
+type RedshiftServer struct {
```
poundifdef, line 11: Each Redshift destination will have a separate credential for the S3 bucket it uses to upload. In addition to the DB config options, we need to include S3 credentials. 

In `Step 2` of the ticket, you can see the additional fields that we need for S3.

poundifdef, line 11: This is missing the `delete_from_s3` configuration option.

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (s *RedshiftServer) CreateEmptyTable(table string) error {
+	sql := fmt.Sprintf("CREATE TABLE IF NOT EXISTS \"%s\" (__row_id BIGINT)", table)
+	_, err := s.conn.Exec(sql)
+	return err
+}
+
+func (s *RedshiftServer) InsertFromNDJsonFile(table string, fileName string) error {
+	var configOptions config.ScratchDataConfig
+	err := cleanenv.ReadConfig(os.Args[1], &configOptions)
```
poundifdef, line 77: Instead of reading the S3 configuration options from a file, we should be using the values from the `RedshiftConfiguration` struct. Every Redshift configuration will have a different S3 config.

poundifdef (changes requested):
Thank you for this. I've left comments for review.

Rutik7066:
Thank @poundifdef for review I am working on it...

Rutik7066:
Hi @poundifdef fixes done. Please check.

Rutik7066:
Hi @poundifdef I just saw @[mohanish2504](https://github.com/mohanish2504) mohanish2504 is aslo working on it. i hope just my work won't be waste.

poundifdef:
Thank you for your comment. I will be moving ahead with PR #120 for this bounty. Your pull request unfortunately has several blocking bugs. For example, the SQL `CREATE TABLE IF NOT EXISTS...` does not work in Redshift because it does not support the `IF NOT EXISTS` clause. 

Thank you again for looking at this issue.

PR #120, opened 2024-03-16
Feat-red_shift_destination

### LINKED ISSUE #117: Redshift Destination

mohanish2504:
/closes #117
/claim #117

Completes all the steps:
1. Implement Redsift interface.
2. Inserts Based on S3 and Copy. 
3. Delete file from S3 once done.
3. Queries (Both JSON and CSV)

Screenshots:
Screenshot from 2024-03-16 20-01-11
Screenshot from 2024-03-16 19-59-37
Screenshot from 2024-03-16 19-59-28


mohanish2504:
@poundifdef this is ready for review

config.yaml
```diff
@@ -24,9 +24,20 @@ database:
   type: static
 
 destinations:
-  - type: duckdb
+  - type: redshift
```
poundifdef, line 27: Please do not modify the default config.yaml. This file is used as the default config if no credentials are provided.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/scratchdata/scratchdata
 
-go 1.22
+go 1.22.0
```
poundifdef, line 3: 1.22.1

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,170 @@
...
+		m[strings.ToLower(name)] = true
+	}
+	if len(m) == 0 {
+		return nil, fmt.Errorf("getGolumnNames: no columns found: %w", err)
+	}
+	return m, nil
+}
+
+func (s *RedshiftServer) createColumns(table string, jsonTypes map[string]string) error {
+	cols, err := s.getGolumnNames(table)
```
poundifdef, line 49: No need for this. Just use IF NOT EXISTS in SQL when creating columns.

pkg/storage/blobstore/s3/s3.go
```diff
@@ -73,3 +88,28 @@ func NewStorage(c map[string]any) (*Storage, error) {
 
 	return q, nil
 }
+
+func NewStorageWithCreds(accessKeyID, secretAccessKey, bucket, region string) (*Storage, error) {
```
poundifdef, line 92: This function should just call `NewStorage()` with the parameters passed in. Do not call `LoadDefaultConfig()`, that is only intended to be used if someone runs the program without any config file at all.

pkg/destinations/redshift/redshift.go
```diff
@@ -0,0 +1,67 @@
...
+		s.Port,
+		s.Database)
+
+	var err error
+	var db *sql.DB
+
+	if db, err = sql.Open("postgres", url); err != nil {
+		return nil, fmt.Errorf("redshift conn error : (%v)", err)
+	}
+	log.Printf("Connecting to Redshift %v", url)
```
poundifdef, line 42: use the `zerolog` package for all logging.

pkg/destinations/redshift/redshift.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"database/sql"
+	"fmt"
+	"log"
+
+	_ "github.com/lib/pq"
+	"github.com/scratchdata/scratchdata/util"
+)
+
+type RedshiftServer struct {
```
poundifdef, line 12: This is missing the `delete_from_s3` setting, see Step 2 of #117

pkg/destinations/redshift/query.go
```diff
@@ -0,0 +1,115 @@
...
+			return fmt.Errorf("failed to scan row values: %v", err)
+		}
+
+		// Populate the JSON object with column values
+		jsonObject := make(map[string]interface{})
+		for i, column := range columns {
+			jsonObject[column] = values[i]
+		}
+
+		jsonObjects = append(jsonObjects, jsonObject)
```
poundifdef, line 44: Do not load the entire dataset in memory. Instead, stream data as it is generated. See Step 2 of #117:

> Data should be streamed to the writer and avoid allocations as much as possible. Do not buffer all data into memory before returning to the client. Do not generate the entire JSON/CSV payload in memory.

pkg/destinations/redshift/redshift.go
```diff
@@ -0,0 +1,67 @@
...
+	Host     string `mapstructure:"redshift_host"`
+	Port     int    `mapstructure:"redshift_port"`
+	Username string `mapstructure:"redshift_user"`
+	Password string `mapstructure:"redshift_password"`
+	Database string `mapstructure:"redshift_database"`
+	Schema   string `mapstructure:"redshift_schema"`
```
poundifdef, line 18: `Schema` is only used in `getColumnNames()`. Should it be used for all queries?

poundifdef (changes requested):

pkg/destinations/redshift/redshift.go
mohanish2504: Then, how are we planning to take schema?

mohanish2504: Yes

pkg/destinations/redshift/query.go
mohanish2504: Noted

pkg/destinations/redshift/redshift.go
mohanish2504: delete_from_s3 will be from config file too?

pkg/destinations/redshift/insert.go
mohanish2504: @poundifdef not supported in aws redshift

mohanish2504: its weird though, it is supported while creating tables, but not altering one.

pkg/destinations/redshift/redshift.go
mohanish2504: @poundifdef isn't queries made by user?
how can we add schema their?

pkg/destinations/redshift/query.go
mohanish2504: Done

pkg/destinations/redshift/query.go
```diff
@@ -0,0 +1,115 @@
...
+			return fmt.Errorf("failed to write CSV row: %v", err)
+		}
+	}
+
+	if err := rows.Err(); err != nil {
+		return fmt.Errorf("failed to iterate over rows: %v", err)
+	}
+
+	return nil
+}
```
mohanish2504, line 115: Write row one by one rather, don;t load all in meonry

pkg/destinations/redshift/redshift.go
poundifdef: Yes. Look at #117. Under "Implement Data Insertion":

> If the delete_from_s3 flag is true, then delete the file from S3 if the copy command has been successful.

pkg/destinations/redshift/redshift.go
poundifdef: Good point.

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,170 @@
...
+	err = s3Client.Upload(filePath, file)
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
+		return err
+	}
+
+	copyCommand := fmt.Sprintf("COPY %s FROM 's3://%s/%s' CREDENTIALS 'aws_access_key_id=%s;aws_secret_access_key=%s' FORMAT AS JSON 'auto'", table, s.S3Bucket, filePath, s.S3AccessKeyId, s.S3SecretAccessKey)
```
poundifdef, line 158: Shouldn't this use `Schema`?

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,170 @@
...
+		case "string":
+			colType = "VARCHAR"
+		default:
+			colType = "VARCHAR"
+		}
+
+		if _, ok := cols[strings.ToLower(colName)]; !ok {
+			log.Printf("Column %s does not exist, creating it", colName)
+			sql := fmt.Sprintf("ALTER TABLE \"%s\" ADD COLUMN \"%s\" %s", table, colName, colType)
+			_, err = s.conn.Exec(sql)
```
poundifdef, line 74: Can we check the type of error here? If two processes attempt to insert data at the same time, then we might find that some of the columns have already been created even when we execute this.

If the error is that the column already exists, then we can ignore the error and continue inserting. However if there is a different error, then we need to stop the insert process.

pkg/destinations/redshift/insert.go
mohanish2504: Yes, good catch.

pkg/destinations/redshift/insert.go
mohanish2504: So what are suggesting is too directly executing the Alter Add Column command?
Then if error is about already column existing, move ahead, else throw error?

poundifdef: Correct

pkg/destinations/redshift/redshift.go
mohanish2504: this is resolved or do you have any idea ?

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+		sql := fmt.Sprintf("ALTER TABLE \"%s\" ADD COLUMN \"%s\" %s", table, colName, colType)
+		_, err := s.conn.Exec(sql)
+		if err != nil {
+			if !strings.Contains(err.Error(), "already exists") {
+				log.Err(err).Msg("createColumns: cannot create column")
+				return err
+			}
+
+		}
+		log.Info().Msgf("Created column %s %s", colName, colType)
```
poundifdef, line 42:
```suggestion
		log.Trace().Str("name", colName).Str("type",colType).Msg("Created column")
```

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+			colType = "VARCHAR"
+		default:
+			colType = "VARCHAR"
+		}
+
+		sql := fmt.Sprintf("ALTER TABLE \"%s\" ADD COLUMN \"%s\" %s", table, colName, colType)
+		_, err := s.conn.Exec(sql)
+		if err != nil {
+			if !strings.Contains(err.Error(), "already exists") {
+				log.Err(err).Msg("createColumns: cannot create column")
```
poundifdef, line 37:
```suggestion
				log.Error().Err(err).Str("column", colName).Msg("createColumns: cannot create column")
```

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+			colType = "BOOLEAN"
+		case "float":
+			colType = "DOUBLE PRECISION"
+		case "string":
+			colType = "VARCHAR"
+		default:
+			colType = "VARCHAR"
+		}
+
+		sql := fmt.Sprintf("ALTER TABLE \"%s\" ADD COLUMN \"%s\" %s", table, colName, colType)
```
poundifdef, line 33: Should this use `Schema`?

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+		_, err := s.conn.Exec(sql)
+		if err != nil {
+			if !strings.Contains(err.Error(), "already exists") {
+				log.Err(err).Msg("createColumns: cannot create column")
+				return err
+			}
+
+		}
+		log.Info().Msgf("Created column %s %s", colName, colType)
+		// if _, ok := cols[strings.ToLower(colName)]; !ok {
```
poundifdef, line 43: Delete

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+		log.Error().Err(err).Str("filename", fileName).Msg("Unable to close file")
+	}
+
+	return nil
+
+}
+
+func (s *RedshiftServer) CreateEmptyTable(table string) error {
+
+	sql := fmt.Sprintf("CREATE TABLE IF NOT EXISTS \"%s\" (__row_id BIGINT)", table)
```
poundifdef, line 79: Should this use `Schema`?

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+
+	}
+
+	params := make(map[string]any)
+
+	params["region"] = s.S3Region
+	params["access_key_id"] = s.S3AccessKeyId
+	params["secret_access_key"] = s.S3SecretAccessKey
+	params["bucket"] = s.S3Bucket
+	params["skipDefaultConfig"] = true
```
poundifdef, line 100: I don't understand the reason for `skipDefaultConfig`. Can you describe why we need this check? Could we simplify the code if we don't use it?

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+		log.Fatal().Err(err).Msg("Unable to create blobstore")
+	}
+
+	file, err := os.Open(filePath)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	absoluteFile, err := filepath.Abs(filePath)
```
poundifdef, line 113: We don't want an absolute file path here. That would result in very confusing paths in S3. I think a path like `<table>/<filename>.json` would be better. Perhaps we could also let the user configure an S3 `prefix` for where to store these temporary files.

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return err
+	}
+	if s.DeleteFromS3 {
+		log.Info().Msgf("Deleting file %s from S3", filePath)
+		err = s3Client.Delete(filePath)
+
+		if err != nil {
+			log.Err(err).Msg("Failed to delete file from S3")
+			return err
```
poundifdef, line 145: If the deletion fails, we should log an error but still return nil since the data has still been uploaded successfully.

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	copyCommand := fmt.Sprintf("COPY %s FROM 's3://%s/%s' CREDENTIALS 'aws_access_key_id=%s;aws_secret_access_key=%s' FORMAT AS JSON 'auto'", s.Schema+"."+table, s.S3Bucket, filePath, s.S3AccessKeyId, s.S3SecretAccessKey)
+
+	_, err = s.conn.Exec(copyCommand)
+	if err != nil {
+		return err
+	}
+	if s.DeleteFromS3 {
+		log.Info().Msgf("Deleting file %s from S3", filePath)
```
poundifdef, line 140:
```suggestion
		log.Trace().Str("file_path",filePath).Msg("Deleting file from S3")
```

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,149 @@
...
+		err = s3Client.Delete(filePath)
+
+		if err != nil {
+			log.Err(err).Msg("Failed to delete file from S3")
```
poundifdef, line 144: Does this work? I think the syntax is supposed to be `log.Error().Err()...`

```suggestion
			log.Error().Err(err).Str("file_path",filePath).Msg("Failed to delete file from S3")
```

pkg/destinations/redshift/query.go
```diff
@@ -0,0 +1,143 @@
...
+			log.Err(err).Msg("failed to scan row values")
+			return err
+		}
+
+		jsonObject := make(map[string]interface{})
+		for i, column := range columns {
+			jsonObject[column] = values[i]
+		}
+
+		jsonData, err := json.Marshal(jsonObject)
```
poundifdef, line 51: Would it make sense to use json.Encoder instead? https://pkg.go.dev/encoding/json#NewEncoder

pkg/destinations/redshift/query.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	values := make([]interface{}, len(columns))
+	valuePtrs := make([]interface{}, len(columns))
+	for i := range columns {
+		valuePtrs[i] = &values[i]
+	}
+
+	// Write column names to the writer
+	_, err = writer.Write([]byte(strings.Join(columns, ",") + "\n"))
```
poundifdef, line 109: Use the csv package instead: https://pkg.go.dev/encoding/csv

pkg/storage/blobstore/s3/s3.go
```diff
@@ -50,22 +52,57 @@ func (s *Storage) Download(path string, w io.WriterAt) error {
 	return nil
 }
 
+func (s *Storage) Delete(path string) error {
+	_, err := s.client.DeleteObject(context.TODO(), &s3.DeleteObjectInput{
+		Bucket: aws.String(s.Bucket),
+		Key:    aws.String(path),
+	})
+
+	if err != nil {
```
poundifdef, line 61:
```suggestion
	return err
```

pkg/storage/blobstore/s3/s3.go
```diff
@@ -50,22 +52,57 @@ func (s *Storage) Download(path string, w io.WriterAt) error {
...
+	return nil
+}
+
 // NewStorage returns a new initialized Storage
 func NewStorage(c map[string]any) (*Storage, error) {
+	var cfg aws.Config
+	var err error
+	skipDefaultConfig := false
 
+	val, ok := c["skipDefaultConfig"]
```
poundifdef, line 74: I don't really understand why we need `skipDefaultConfig`.

poundifdef (changes requested):

pkg/storage/blobstore/s3/s3.go
mohanish2504: Won't it fail if it doens't found any config from env/context

does this suggested change works?.

Screenshot from 2024-03-16 23-42-51

poundifdef: Perhaps we should replace `LoadDefaultConfig()` with `cfg := aws.Config{}`. We never want to load anything from the environment. We only ever want to load data from this configuration file.

We can create a ticket to be able configuration options from environment variables or credentials files.

Thank you for clarifying.

mohanish2504:
@poundifdef are we suppose to use? `log.Err(err)` or `log.Error().Err(err)` every where because I see multiple other places same code has been implemented?

poundifdef:
@mohanish2504 I just checked - I did not realize that `log.Err()` and `log.Error().Err()` are equivalent. You can disregard that comment.

pkg/destinations/redshift/insert.go
poundifdef: While we don't need `Error().Err()`, we still want to avoid `Msgf()` in favor of `Str("column", colName)`. This will let us index logs easier.

mohanish2504:
my concern is that it is also being used in other places other than redshift, so I got confuse which one to use.
for now may be I should keep it as it is? or make the changes everywhere

mohanish2504:
Here are the differences
`log.Err()`
Screenshot from 2024-03-17 00-48-44

`log.Error().Err()`
Screenshot from 2024-03-17 00-47-54

pkg/destinations/redshift/insert.go
mohanish2504: Noted

poundifdef:
For this PR specifically, please use `log.Error().Err()`. No need to fix this in other places in the code.

mohanish2504:
@poundifdef I tried merging latest changes, but it throws some error for auth.

Shall I still add the merge?

poundifdef:
What is the error? You should be able to run:

```
go run .
```

And then 

```
curl 'http://localhost:8080/api/data/query?query=select%201&api_key=local'
```

mohanish2504:
> What is the error? You should be able to run:
> 
> ```
> go run .
> ```
> 
> And then
> 
> ```
> curl 'http://localhost:8080/api/data/query?query=select%201&api_key=local'
> ```

Error:

`
01:23:27 DBG scratchdata.go:106 > Starting Scratch Data
01:23:27 DBG api.go:86 > Starting API port=8080
01:23:27 DBG workers.go:141 > Starting Workers
01:23:27 DBG workers.go:26 > Starting worker thread=0
2024/03/17 01:23:29 http: panic serving 127.0.0.1:45350: runtime error: invalid memory address or nil pointer dereference
goroutine 7 [running]:
net/http.(*conn).serve.func1()
        /home/mak/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.1.linux-amd64/src/net/http/server.go:1898 +0xbe
panic({0x2397be0?, 0x2e60dc0?})
        /home/mak/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.1.linux-amd64/src/runtime/panic.go:770 +0x132
github.com/scratchdata/scratchdata/pkg/api.(*ScratchDataAPIStruct).AuthMiddleware.func1({0x25f7ac8, 0xc0000b41c0}, 0xc0000ba480)
        /home/mak/opensource/scratchdata/pkg/api/api.go:54 +0x74
net/http.HandlerFunc.ServeHTTP(0x25f8630?, {0x25f7ac8?, 0xc0000b41c0?}, 0x2e60ed0?)
        /home/mak/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.1.linux-amd64/src/net/http/server.go:2166 +0x29
github.com/go-chi/chi/v5.(*Mux).ServeHTTP(0xc00014c600, {0x25f7ac8, 0xc0000b41c0}, 0xc0000ba120)
        /home/mak/go/pkg/mod/github.com/go-chi/chi/v5@v5.0.12/mux.go:90 +0x2ee
net/http.serverHandler.ServeHTTP({0xc0000b63c0?}, {0x25f7ac8?, 0xc0000b41c0?}, 0x6?)
        /home/mak/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.1.linux-amd64/src/net/http/server.go:3137 +0x8e
net/http.(*conn).serve(0xc0000d4000, {0x25f85f8, 0xc0000b62d0})
        /home/mak/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.1.linux-amd64/src/net/http/server.go:2039 +0x5e8
created by net/http.(*Server).Serve in goroutine 5
        /home/mak/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.1.linux-amd64/src/net/http/server.go:3285 +0x4b4
`

poundifdef:
Obviously do not merge that. If you need to revert back to `config.LoadDefaultConfig(context.TODO())` then that is fine.

poundifdef:
You probably need to create a separate `config.yaml` file and pass that in: `go run . config.yaml`. That local config file should have a full redshift configuration.

mohanish2504:
> config.LoadDefaultConfig(context.TODO())

I don't think so this is related to this change. It is working with this. I have already checked and pushed.

Our merge conflict was on imports.

mohanish2504:
> You probably need to create a separate `config.yaml` file and pass that in: `go run . config.yaml`. That local config file should have a full redshift configuration.

Already have this.

poundifdef:
Sorry about the conflicts. I think they should be easy for you to fix, looks like it is just one or two imports.

mohanish2504:
@poundifdef already merged it.

poundifdef:
Okay. Let me know when you're ready for me to test the code. Thank you for all of the work you've been putting into this feature.

mohanish2504:
@poundifdef sure, how do we want to test the code? you will be needing something from my end?

mohanish2504:
destinations:
  - type: redshift
    name: Local DuckDB file
    api_keys:
      - local 
    settings:
  
      redshift_host: .redshift-serverless.amazonaws.com
      redshift_port: 5439 
      redshift_user: user
      redshift_password: psswd
      redshift_database: dev
      redshift_schema: public

      s3_region: region
      s3_access_key_id: abc1234
      s3_secret_access_key: abc1234
      s3_bucket: bucket
      
      for your reference

poundifdef (changes requested):
I've tested this locally with my own AWS account and everything looks good! Just found one bug. If you fix it then I'm happy to merge the code and complete the bounty. Thank you!

pkg/destinations/redshift/insert.go
```diff
@@ -0,0 +1,141 @@
...
+			colType = "BOOLEAN"
+		case "float":
+			colType = "DOUBLE PRECISION"
+		case "string":
+			colType = "VARCHAR"
+		default:
+			colType = "VARCHAR"
+		}
+
+		sql := fmt.Sprintf("ALTER TABLE \"%s\" ADD COLUMN \"%s\" %s", s.Schema+"."+table, colName, colType)
```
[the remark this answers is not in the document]
mohanish2504: Yes, just a second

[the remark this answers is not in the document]
mohanish2504: Done

mohanish2504:
Befor you close the issue @poundifdef.

Thanks for your patience, looking forward to close more such critical/challenging bugs (bounty/non-bounty).

Please tag me if you find something interesting for me.

poundifdef (approved):

PR #122, opened 2024-03-16
feat: sharable URLs for querying data

### LINKED ISSUE #116: Sharable URLs for Querying Data
Today, you can query the API by calling an endpoint like so:

```
http://localhost:8080/api/data/query?api_key=local&query=select * from events" 
```

We want to be able to create "sharable URLs." This is how it should work:

## New Endpoint

- [ ] Create a new POST endpoint for creating a query. It should be `POST /api/data/query/share?api_key=local`. The query should be part of the POST request body.
- [ ] The POST request body should also take a duration, which is the number of seconds for the link to be valid for.
- [ ] This endpoint should be authenticated via API key
- [ ] The return value should be a new UUID which represents the query.

## Save query to cache

- [ ] We have a cache [interface](https://github.com/scratchdata/scratchdata/blob/main/pkg/storage/cache/cache.go) which has no implementation. Implement an in-memory cache. You may use a library such as https://github.com/eko/gocache
- [ ] The cache key should be a UUID. 
- [ ] The cache value should store the database ID and query. It should also respect the expiration directive.

## Endpoint to Query Data

- [ ] Create a new endpoint: `GET /share/<uuid>/data.<format>`
- [ ] This endpoint should not be authenticated
- [ ] It should run the query for the given `<uuid>` and return the format specified by `<format>`. This will either be `csv` or `json`. 
- [ ] You should refactor the existing [Select endpoint](https://github.com/scratchdata/scratchdata/blob/main/pkg/api/data.go#L14) in order to reuse code for streaming data to the client

mehulmathur16:
/claim #116
/fixes #116

https://github.com/scratchdata/scratchdata/assets/64700961/5224594d-1da1-45da-b74b-dd0d7f570f78

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -56,6 +56,11 @@ type ScratchDataAPI interface {
 	AuthGetDatabaseID(context.Context) int64
 }
 
+type CachedQueryData struct {
+    Query      string
+    DatabaseID int64
+}
+
 func (a *ScratchDataAPIStruct) AuthMiddleware(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		apiKey := r.URL.Query().Get("api_key")
@@ -77,13 +82,6 @@ func (a *ScratchDataAPIStruct) AuthGetDatabaseID(ctx context.Context) int64 {
 }
 
 func (a *ScratchDataAPIStruct) CreateQuery(w http.ResponseWriter, r *http.Request) {
-	apiKey := r.URL.Query().Get("api_key")
-	if apiKey != "local" { // Validate API key
-		w.WriteHeader(http.StatusUnauthorized)
-		w.Write([]byte("Unauthorized"))
-		return
-	}
-
 	var requestBody struct {
 		Query    string `json:"query"`
 		Duration int    `json:"duration"` // Duration in seconds
@@ -102,12 +100,24 @@ func (a *ScratchDataAPIStruct) CreateQuery(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
+	cachedQueryData := CachedQueryData{
+        Query:      requestBody.Query,
+        DatabaseID: a.AuthGetDatabaseID(r.Context()),
+    }
+    cachedQueryDataBytes, err := json.Marshal(cachedQueryData)
+
+    if err != nil {
+        w.WriteHeader(http.StatusInternalServerError)
+        w.Write([]byte("Failed to marshal data"))
+        return
+    }
+
 	// Generate a new UUID
 	queryUUID := uuid.New()
 
 	// Store the query and its expiration time
 	queryExpiration := time.Duration(requestBody.Duration)
-	a.cache.Set(queryUUID.String(), []byte(requestBody.Query), &queryExpiration)
+	a.cache.Set(queryUUID.String(), cachedQueryDataBytes, &queryExpiration)
 
 	// Return the UUID representing the query
 	w.Header().Set("Content-Type", "application/json")
@@ -140,17 +150,23 @@ func (a *ScratchDataAPIStruct) ShareData(w http.ResponseWriter, r *http.Request)
 	format := chi.URLParam(r, "format")
 
 	// Retrieve query from cache using UUID
-	query, found := a.cache.Get(queryUUID)
+	cachedQueryDataBytes, found := a.cache.Get(queryUUID)
 	if !found {
 		http.Error(w, "Query not found", http.StatusNotFound)
 		return
 	}
 
+	var cachedQueryData CachedQueryData
+    if err := json.Unmarshal(cachedQueryDataBytes, &cachedQueryData); err != nil {
+        http.Error(w, "Failed to unmarshal data", http.StatusInternalServerError)
+        return
+    }
+
 	// Convert query to string
-	queryStr := string(query)
+	queryStr := string(cachedQueryData.Query)
 
 	// Execute query and stream data
-	databaseID := a.AuthGetDatabaseID(r.Context())
+	databaseID := cachedQueryData.DatabaseID
 	if err := a.executeQueryAndStreamData(w, r.Context(), queryStr, databaseID, format); err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -34,12 +34,23 @@ func NewScratchDataAPI(storageServices *models.StorageServices, destinationManag
 		return nil, err
 	}
 
+	cacheConfig := config.Cache{
+		Type:     "memory",
+		Settings: map[string]interface{}{},
+	}
+
+	// Call NewCache with the configuration
+	cacheInstance, err := cache.NewCache(cacheConfig)
+	if err != nil {
+		return nil, err
+	}
+
 	rc := ScratchDataAPIStruct{
 		storageServices:    storageServices,
 		destinationManager: destinationManager,
 		dataSink:           dataSink,
 		snow:               snow,
-		cache:              *cache.NewCache(),
+		cache:              cacheInstance,
 	}
 
 	return &rc, nil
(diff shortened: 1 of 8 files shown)


mehulmathur16:
Hi @poundifdef, can you please review this PR. I have attached a screen recording of the feature for your reference.

pkg/api/api.go
```diff
@@ -68,6 +76,86 @@ func (a *ScratchDataAPIStruct) AuthGetDatabaseID(ctx context.Context) int64 {
 	return ctx.Value("databaseId").(int64)
 }
 
+func (a *ScratchDataAPIStruct) CreateQuery(w http.ResponseWriter, r *http.Request) {
+	apiKey := r.URL.Query().Get("api_key")
```
poundifdef, line 80: You do not need to do any API key validation here because it is all done in`AuthMiddleware`

pkg/api/api.go
```diff
@@ -68,6 +76,86 @@ func (a *ScratchDataAPIStruct) AuthGetDatabaseID(ctx context.Context) int64 {
...
+		w.Write([]byte("Query cannot be empty"))
+		return
+	}
+
+	// Generate a new UUID
+	queryUUID := uuid.New()
+
+	// Store the query and its expiration time
+	queryExpiration := time.Duration(requestBody.Duration)
+	a.cache.Set(queryUUID.String(), []byte(requestBody.Query), &queryExpiration)
```
poundifdef, line 110: The cache needs to store both the Query and the `databaseId`. See #116 under "Save query to cache"

> The cache value should **store the database ID and query**. It should also respect the expiration directive.

pkg/api/api.go
```diff
@@ -68,6 +76,86 @@ func (a *ScratchDataAPIStruct) AuthGetDatabaseID(ctx context.Context) int64 {
...
+	if !found {
+		http.Error(w, "Query not found", http.StatusNotFound)
+		return
+	}
+
+	// Convert query to string
+	queryStr := string(query)
+
+	// Execute query and stream data
+	databaseID := a.AuthGetDatabaseID(r.Context())
```
poundifdef, line 153: `databaseID` should be retrieved from the cache. It should not come from `r.Context()`. That is because the query is specific to a database. With the current code, a saved query could be run against a different database, which would be an error.

pkg/api/api.go
```diff
@@ -76,6 +164,8 @@ func CreateMux(apiFunctions ScratchDataAPI) *chi.Mux {
 	api.Post("/data/insert/{table}", apiFunctions.Insert)
 	api.Get("/data/query", apiFunctions.Select)
 	api.Post("/data/query", apiFunctions.Select)
+	api.Post("/data/query/share", apiFunctions.CreateQuery)        // New endpoint for creating a query
+	api.Get("/share/{uuid}/data.{format}", apiFunctions.ShareData) // New endpoint for sharing data
```
poundifdef, line 168: This path should be `/share/{uuid}...`. However, this route is `/api/share...` which is not what we want.

```suggestion
	r.Get("/share/{uuid}/data.{format}", apiFunctions.ShareData) // New endpoint for sharing data
```

poundifdef: In order to make this work, perhaps the `r.Use(apiFunctions.AuthMiddleware)` should instead be `api.User(...)`.

pkg/api/api.go
```diff
@@ -53,7 +61,7 @@ func (a *ScratchDataAPIStruct) AuthMiddleware(next http.Handler) http.Handler {
 		apiKey := r.URL.Query().Get("api_key")
 		keyDetails, err := a.storageServices.Database.GetAPIKeyDetails(apiKey)
 
-		if err != nil {
+		if err != nil && !strings.HasPrefix(r.URL.Path, "/api/share") {
```
poundifdef, line 64: Instead of hard-coding this path, look at my suggestions on line 168.

pkg/api/api.go
```diff
@@ -68,6 +76,86 @@ func (a *ScratchDataAPIStruct) AuthGetDatabaseID(ctx context.Context) int64 {
...
+	w.Header().Set("Content-Type", "application/json")
+	response := struct {
+		QueryUUID string `json:"query_uuid"`
+	}{
+		QueryUUID: queryUUID.String(),
+	}
+	json.NewEncoder(w).Encode(response)
+}
+
+func (a *ScratchDataAPIStruct) executeQueryAndStreamData(w http.ResponseWriter, ctx context.Context, query string, databaseID int64, format string) error {
```
poundifdef, line 122: Thank you for creating a new function. Can you update the `Select()` function to also use this? See #116 under "Endpoint to Query Data"

> You should refactor the existing [Select endpoint](https://github.com/scratchdata/scratchdata/blob/main/pkg/api/data.go#L14) in order to reuse code for streaming data to the client

pkg/storage/cache/cache.go
```diff
@@ -1,8 +1,37 @@
 package cache
 
-import "time"
+import (
+	"time"
 
-type Cache interface {
```
poundifdef, line 5: Do not remove this interface. Keep the interface as it is, and create a new package which implements the interface. This is what we do here: https://github.com/scratchdata/scratchdata/tree/main/pkg/storage/queue

poundifdef (changes requested):
Thank you. This looks good, just a few small changes.

pkg/storage/cache/cache.go
mehulmathur16: Can you please elaborate more on the structure of storage/cache directory ?

mehulmathur16:
@poundifdef I've resolved the comments except the cache pkg. I need more clarity on it.

Also, Can you please assign me this issue ?

pkg/storage/cache/cache.go
poundifdef: The `storage/cache` directory might have many implementations of a cache (for example, in-memory and redis.) We use a go interface in order to specify how each cache should behave.

Create a new folder called `memory` under `storage/cache` where you have your implementation with your `Cache` struct. Then, any other place in the code which uses a cache can pass the interface type, regardless of the underlying implementation.

You may look at the `storage/queue` directory structure for a template on how to do this.

pkg/api/api.go
```diff
@@ -68,14 +80,91 @@ func (a *ScratchDataAPIStruct) AuthGetDatabaseID(ctx context.Context) int64 {
...
+	}
+
+	var cachedQueryData CachedQueryData
+	if err := json.Unmarshal(cachedQueryDataBytes, &cachedQueryData); err != nil {
+		http.Error(w, "Failed to unmarshal data", http.StatusInternalServerError)
+		return
+	}
+
+	// Convert query to string
+	queryStr := string(cachedQueryData.Query)
```
poundifdef, line 149: `Query` is already a string, right?

pkg/storage/cache/cache.go
mehulmathur16: @poundifdef I have created the directories. Can you please tell how to call new cache now ?

image

mehulmathur16:
@poundifdef I have added the cache pkg. Please review once.

pkg/storage/cache/cache.go
mehulmathur16: Thanks! This is resolved.

poundifdef (changes requested):
This looks good. Just one small change and then I'll merge. Thank you!

mehulmathur16:
Hi @poundifdef, I have made the requested changes. Please review.

pkg/api/api.go
```diff
@@ -35,6 +39,7 @@ func NewScratchDataAPI(storageServices *models.StorageServices, destinationManag
 		destinationManager: destinationManager,
 		dataSink:           dataSink,
 		snow:               snow,
+		cache:              storageServices.Cache,
```
poundifdef, line 42: This is not needed. Any functions that need the cache can just call `storageService.Cache`

mehulmathur16:
Hi @poundifdef, please review.

poundifdef (approved):

poundifdef:
This looks good. Thank you.

mehulmathur16:
> This looks good. Thank you.

That's great! Thankyou for the opportunity.
Looking forward for more contributions.

PR #123, opened 2024-03-17
google bigquery implementation complete

### LINKED ISSUE #115: BigQuery Destination
We want to be able to read and write data from BigQuery in bulk. Here are the steps to accomplish that:

## Step 1: Create new BigQuery Destination interface

- [ ] Add a new package in [pkg/destinations](https://github.com/scratchdata/scratchdata/tree/main/pkg/destinations) called `bigquery`
- [ ] Update [destinations.go](https://github.com/scratchdata/scratchdata/blob/main/pkg/destinations/destinations.go#L69-L74) to be able to connect to the new bigquery destination.
- [ ] Implement the functions in the [Destination interface](https://github.com/scratchdata/scratchdata/blob/main/pkg/destinations/destinations.go#L20) for the new `bigquery` package. For this first step, you may stub these functions (`panic("not implemented")`)

## Step 2: Implement Queries from BigQuery

Set up configuration parameters for BigQuery.

``` yaml
databases:
  - type: bigquery
    api_keys:
      - test_api_key
    settings:
      json_credentials: >
        {
          "type": "service_account",
          "project_id": "example-123",
          "private_key_id": "...",
          "private_key": "-----BEGIN PRIVATE KEY-----...",
          ... 
        }
```

- [ ] The `Close()` function should clean up the BigQuery connection, and we should also close any open handlers related to executing the query.

- [ ] Implement the `CreateColumns()` function. This should read the input file and alter the BigQuery table to add columns if they do not exist. [Here](https://github.com/scratchdata/scratchdata/blob/main/pkg/destinations/duckdb/insert.go#L79) is a reference for how to do this.

## Implement Data Insertion

We want to bulk upload data to BigQuery from the input file. This means imlementing the `InsertFromNDJsonFile()` function.

- [ ] Data should be streamed to BigQuery. Do not load the entire data set into memory when uploading to the database.
- [ ] Implement this with the [Load](https://cloud.google.com/bigquery/docs/batch-loading-data#loading_data_from_local_files) API. The input type is newline-delimited JSON.

mohanish2504:
/claim #115
/closes #115

- [x] QueryCSV and QueryJSON
- [x] InsertFromNDJson: Uploads to CSV to GCS and then to Bigquery
- [x] Added GCS storage provider (gcs and s3 are interoperable, tried with but didn't work)
- [x] Optional GCS delete 
- [x] Optional GCS prefix

@poundifdef  Some thigs to take care:
- Both query fromats require's user to query with `dataset.table`, may be need to somewhere in docs

Screenshots:
Screenshot from 2024-03-17 17-59-04
Screenshot from 2024-03-17 17-56-39
Screenshot from 2024-03-17 18-00-04
Screenshot from 2024-03-17 18-17-10


mohanish2504:
@poundifdef this is for ready review

pkg/destinations/bigquery/bigquery.go
```diff
@@ -0,0 +1,98 @@
...
+
+	credentials, err := google.CredentialsFromJSON(ctx, []byte(s.CredentialsJsonString), bigquery.Scope)
+	if err != nil {
+		log.Error().Err(err).Msg("bigquery credentials error")
+		return nil, err
+	}
+	var projectId string
+	var ok bool
+
+	projectId, ok = credentialsJson["project_id"].(string)
```
poundifdef, line 58:
```suggestion
	projectId = credentials.ProjectID
```

pkg/destinations/bigquery/bigquery.go
```diff
@@ -0,0 +1,98 @@
...
+	err := json.Unmarshal([]byte(credentialsString), &credentials)
+	if err != nil {
+		return nil, err
+	}
+	return credentials, nil
+}
+
+func openConn(s *BigQueryServer) (*bigquery.Client, error) {
+	ctx := context.Background()
+	credentialsJson, err := parseCredentials(s.CredentialsJsonString)
```
poundifdef, line 44: You don't need this. You can get the ProjectID from `google.CredentialsFromJSON()`

https://pkg.go.dev/golang.org/x/oauth2/google#Credentials

pkg/destinations/bigquery/bigquery.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, err
+	}
+
+	s.Credentials = credentials
+	s.ProjectId = &projectId
+
+	// at this point it is apparent that credential are present and valid
+	privateKeyId, _ := credentialsJson["private_key_id"].(string)
+	privateKey, _ := credentialsJson["private_key"].(string)
+	s.GCSAccessKeyId = &privateKeyId
```
poundifdef, line 77: No need for these fields since they are never used

pkg/destinations/bigquery/insert.go
```diff
@@ -0,0 +1,271 @@
...
+
+	columns := make([]string, 0)
+	for _, field := range meta.Schema {
+		columns = append(columns, field.Name)
+	}
+
+	return columns, nil
+}
+
+func (s *BigQueryServer) ConvertNDJSONToCSVAndUpload(table string, filePath string, csvFilePath string) error {
```
poundifdef, line 117: Why convert to CSV first? Doesn't BigQuery support loading JSON? https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-json

pkg/destinations/bigquery/query.go
```diff
@@ -0,0 +1,129 @@
...
+
+	columns := make([]string, 0)
+	columnsRow := make(map[string]bigquery.Value)
+	if err := columnsItr.Next(&columnsRow); err != nil {
+		log.Error().Err(err).Msg("error retrieving columns")
+		return err
+	}
+
+	//column orders are not ensured in bigquery
+	columns = append(columns, "__row_id")
```
poundifdef, line 79: We don't care about making the __row_id the first column.

mohanish2504: Okay then removing!

pkg/destinations/bigquery/insert.go
mohanish2504: I didn't read ndjson here! Thought only accepts json, changing .... .

pkg/destinations/bigquery/bigquery.go
mohanish2504: Yep

pkg/storage/blobstore/gcs/gcs.go
```diff
@@ -0,0 +1,94 @@
...
+		return err
+	}
+	defer rc.Close()
+
+	objectSize := rc.Attrs.Size
+
+	chunkSize := 1 * 1024 * 1024 // 1MB chunk
+
+	var offset int64 = 0
+	for {
```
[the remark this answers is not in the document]
mohanish2504: io.Copy supports write, doesnt support WirteAt. 
Hod do you usually do this?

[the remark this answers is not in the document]
poundifdef: I think you can change the interface to `*os.File`. S3 uses `WriterAt` and GCS uses `Writer` so I think `*os.File` should cover both.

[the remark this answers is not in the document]
mohanish2504: Yokay

pkg/destinations/bigquery/insert.go
mohanish2504: fixed

pkg/storage/blobstore/gcs/gcs.go
```diff
@@ -0,0 +1,94 @@
...
+
+		buf := make([]byte, chunkSize)
```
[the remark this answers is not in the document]
mohanish2504: using io.Copy now

pkg/storage/blobstore/gcs/gcs.go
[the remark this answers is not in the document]
mohanish2504: Done

pkg/destinations/bigquery/bigquery.go
```diff
@@ -0,0 +1,64 @@
...
+	DatasetID             string `mapstructure:"bigquery_dataset_id"`
+	CredentialsJsonString string `mapstructure:"bigquery_credentials_json"`
+
+	// this will be implemented during runtime
+	ProjectId *string
+
+	GCSBucketName string `mapstructure:"gcs_bucket_name"`
+	GCSFilePrefix string `mapstructure:"gcs_file_prefix"`
+
+	DeleteFromGCS bool `mapstructure:"delete_from_gcs"`
```
poundifdef, line 23: This flag is not being used anywhere.

pkg/destinations/bigquery/bigquery.go
```diff
@@ -0,0 +1,64 @@
...
+
+	"cloud.google.com/go/bigquery"
+	"github.com/rs/zerolog/log"
+	"github.com/scratchdata/scratchdata/util"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/option"
+)
+
+type BigQueryServer struct {
+	DatasetID             string `mapstructure:"bigquery_dataset_id"`
```
poundifdef, line 14: Let's remove this and have the user specify a full table (`dataset_name.table_name`) when they call the POST endpoint. If the dataset doesn't exist, we should just create it for them.

pkg/destinations/bigquery/bigquery.go
```diff
@@ -0,0 +1,64 @@
...
+	CredentialsJsonString string `mapstructure:"bigquery_credentials_json"`
```
poundifdef, line 15:
```suggestion
	CredentialsJsonString string `mapstructure:"credentials_json"`
```

The credentials are used for both GCS and BigQuery

pkg/destinations/bigquery/insert.go
```diff
@@ -0,0 +1,204 @@
...
+		gcsFilePath = s.GCSFilePrefix + "/"
+	}
+	gcsFilePath += table + "/" + filepath.Base(filePath)
+
+	log.Info().Msg("Uploading file to GCS .....")
+	err = s.uploadFileToGCS(table, filePath, gcsFilePath, client)
+	if err != nil {
+		return err
+	}
+	log.Info().Msg("Uploaded!")
```
poundifdef, line 136:
```suggestion
	log.Info().Str("gcs_file", gcsFilePath).Msg("Uploaded to GCS")
```

In general, for logs, it is helpful to have some sort of identifier. For example, include the name of the file that we've uploaded. Otherwise it is impossible to look at a log and know what was uploaded. Please revise the other other log statements to try and add identifiers when it makes sense.

pkg/destinations/bigquery/insert.go
```diff
@@ -0,0 +1,204 @@
...
+		return err
+	}
+
+	gcsFilePath := ""
+	if s.GCSFilePrefix != "" {
+		gcsFilePath = s.GCSFilePrefix + "/"
+	}
+	gcsFilePath += table + "/" + filepath.Base(filePath)
+
+	log.Info().Msg("Uploading file to GCS .....")
```
poundifdef, line 131:
```suggestion
	log.Info().Msg("Uploading file to GCS")
```

pkg/destinations/bigquery/insert.go
```diff
@@ -0,0 +1,204 @@
...
+	err = s.uploadFileToGCS(table, filePath, gcsFilePath, client)
+	if err != nil {
+		return err
+	}
+	log.Info().Msg("Uploaded!")
+
+	log.Info().Msg("Streaming data to BigQuery .....")
```
poundifdef, line 138:
```suggestion
	log.Info().Msg("Streaming data to BigQuery")
```

pkg/destinations/bigquery/insert.go
```diff
@@ -0,0 +1,204 @@
...
+
+	gcsRef := bigquery.NewGCSReference(location)
+	gcsRef.SourceFormat = bigquery.JSON
+
+	loader := tableRef.LoaderFrom(gcsRef)
+	loader.WriteDisposition = bigquery.WriteAppend
+
+	job, err := loader.Run(ctx)
+	if err != nil {
+		log.Error().Err(err).Msg("Failed to run BigQuery loader job")
```
poundifdef, line 179: As before, it would be helpful to have some information, such as the table and gcsFilePath, so when we are debugging we can see exactly what failed.

poundifdef (changes requested):
Thank you - I've tested this and it is almost ready to go. Just a few small changes.

pkg/destinations/bigquery/bigquery.go
mohanish2504: Not sure.

Streaming relies on it, other parts are coverable through sql, not sure how to make this query
 
Screenshot from 2024-03-18 22-52-46

poundifdef: How about parsing the `table` param? `strings.Split(table, ".")`

mohanish2504: Ignore, got it

mohanish2504: The last link works

pkg/destinations/bigquery/insert.go
mohanish2504: Thanks for feedback, I have very weird style of logging.
Looking forward to improve.

mohanish2504:
@poundifdef added changes

poundifdef:
This branch does not work any longer. 

```
curl -X POST 'http://localhost:8080/api/data/insert/my_dataset.t?api_key=bq' --data '{"x": "y"}'
```

Gives this error:

```
14:41:55 ERR insert.go:23 > CreateEmptyTable: failed to create Table error="googleapi: Error 404: Not found: Dataset scratch-data-410814:my_dataset was not found in location US, notFound" query="CREATE TABLE IF NOT EXISTS my_dataset.t (__row_id BIGINT)"
14:41:55 ERR workers.go:41 > Unable to process message error="googleapi: Error 404: Not found: Dataset scratch-data-410814:my_dataset was not found in location US, notFound" thread=0
```

You probably want to use this api, right? https://cloud.google.com/bigquery/docs/datasets#go

mohanish2504:
So split and create the dataset first?

poundifdef:
> So split and create the dataset first?

Yes. The goal is for the user to do as little configuration as possible, so if we can automate creating a dataset then that is what I'd like to do.

mohanish2504:
I might need location parameter, take it from user in config?

poundifdef:
> I will need location parameter, take it from user in config?

Yes

poundifdef:
> How about get query here? What if dataset and table not there? Create and return empty? Most likely user won't be querying such table.

If a user tries to query and the dataset or table does not exist, then we should just return the error that BigQuery throws. We only automatically create tables when the user inserts data.

mohanish2504:
> > How about get query here? What if dataset and table not there? Create and return empty? Most likely user won't be querying such table.
> 
> If a user tries to query and the dataset or table does not exist, then we should just return the error that BigQuery throws. We only automatically create tables when the user inserts data.

That is exactly for current case, pushing.

mohanish2504:
@poundifdef pushed the changes

poundifdef:
Thank you for your persistence. I still get an error when running the same test as above:

```
curl -X POST 'http://localhost:8080/api/data/insert/my_dataset.t?api_key=bq' --data '{"x": "y"}'
```

Here is the error:

```
17:02:36 TRC types.go:59 > column_type_counts={"__row_id":{"int":1},"x":{"string":1}}
17:02:36 TRC types.go:86 > column_types={"__row_id":"int","x":"string"}
17:02:37 INF insert.go:119 > Uploading file to GCS 
17:02:38 INF insert.go:125 > Uploaded file to GCS gcs_file=b/my_dataset.t2/1_my_dataset.t2_1769831789700034560.ndjson.ndjson
17:02:38 INF insert.go:127 > Streaming data to BigQuery
17:02:39 ERR insert.go:172 > StreamDataToBigQuery: failed to stream data to BigQuery error="googleapi: Error 400: Invalid schema update. Field x has changed type from STRING to BOOLEAN, invalid" query="LOAD DATA INTO my_dataset.t2 FROM FILES ( format = 'JSON', uris = ['gs://scratchdata-test-bucket/b/my_dataset.t2/1_my_dataset.t2_1769831789700034560.ndjson.ndjson'] )"
17:02:39 ERR insert.go:130 > Failed to stream data to BigQuery error="googleapi: Error 400: Invalid schema update. Field x has changed type from STRING to BOOLEAN, invalid"
17:02:39 ERR insert.go:184 > Failed to upload and stream data to BigQuery error="googleapi: Error 400: Invalid schema update. Field x has changed type from STRING to BOOLEAN, invalid" file=data/worker/1_my_dataset.t2_1769831789700034560.ndjson.ndjson table=my_dataset.t2
17:02:39 ERR workers.go:41 > Unable to process message error="googleapi: Error 400: Invalid schema update. Field x has changed type from STRING to BOOLEAN, invalid" thread=0
```

I believe this is because BigQuery's `autodetect` is converting the string `"y"` into a boolean before inserting.

mohanish2504:
@poundifdef got chance to test this?

PR #128, opened 2024-03-19
feat: use aws secret manager

### LINKED ISSUE #126: Use a vault for DB credentials

ezhil56x:
Fixes #126 
/claim #126


ezhil56x:
How can I import local packages as the part I'm working on is not already available in GitHub?

poundifdef:
Because this code doesn’t compile, I will close this pr for now.

PR #133, opened 2024-03-23
AWS Secret Manager for DB credentials

### LINKED ISSUE #126: Use a vault for DB credentials

mehulmathur16:
/claim #126
/fixes #126

Hi @poundifdef, I have tested out the memory vault & it is working as expected. In case of aws vault, I do not have an aws account, so wasn't able to test it out. Let me know if you face any issues.

Additionally, I have added Region to the aws config.

### DIFF
--- a/pkg/destinations/destinations.go
+++ b/pkg/destinations/destinations.go
@@ -1,6 +1,7 @@
 package destinations
 
 import (
+	"encoding/json"
 	"errors"
 	"io"
 	"strconv"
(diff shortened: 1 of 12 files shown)


poundifdef:
Can you create an AWS account and test it? If you can confirm that it works then I’m happy to review.

pkg/destinations/destinations.go
```diff
@@ -56,13 +66,21 @@ func (m *DestinationManager) Destination(databaseID int64) (Destination, error)
...
 		var dest Destination
+		var creds Creds
 
 		dest, ok := m.pool[databaseID]
 		if ok {
 			return dest, nil
 		}
 
-		creds, err := m.storage.Database.GetDestinationCredentials(databaseID)
+		jsonCreds, err := m.storage.Vault.GetCredential(strconv.Itoa(int(databaseID)))
```
mehulmathur16, line 76: @poundifdef in which format should the credentials be stored on aws secrets manager ?
Also, should this logic be different when reading from different vaults. As the memory vault returns a json string consisting of all destination details, which is here decoded to Credentials type.

mehulmathur16:
Sure 👍

mehulmathur16:
@poundifdef This branch is too old from main. So creating PR from different branch & closing this.

PR #134, opened 2024-03-23
implement pause and resume data ingest

### LINKED ISSUE #131: Pause and resume data ingest
Today, when data is queued, we immediately consume it and upload to the database. However, we want the ability to pause and resume workers for a specific table, or for all tables in a destination. This is useful because sometimes we perform database migrations and need to pause one table.

## Step 1: One queue per table

- [ ] Instead of a single queue for all data, we should have a separate queue for each table where we insert data
- [ ] If the queue doesn't exist, we should automatically create it
- [ ] You'll need to refactor the [Queue interface](https://github.com/scratchdata/scratchdata/blob/main/pkg/storage/queue/queue.go).
- [ ] [Workers](https://github.com/scratchdata/scratchdata/blob/main/pkg/workers/workers.go#L29) will need to be refactored to consume from multiple queues

This doesn't have to be done with multiple queues if there are other ways (ie, message groups) that would work.

## Step 2: API Endpoint to pause queue for single table and entire destination

- [ ] We want an API endpoint to let us pause+unpause the queue for a single table
- [ ] That endpoint should also let us pause+unpause the queue for all tables in a destination
- [ ] This endpoint will be authenticated
- [ ] The pause status should be stored in the [Database](https://github.com/scratchdata/scratchdata/blob/main/pkg/storage/database/database.go). You'll want to add methods to that interface.

# Step 3: Workers should respect the pause status

- [ ] Now that the workers understand how to use multiple queues, and the database stores the pause status for each destination and table, workers should respect the pause status and only consume from queues that are active.

breadchris:
Closes #131 

Currently, a new queue is made for every table that has items to process. In the case of sqs, this will not work properly unless a new queue is made for every table. An alternative approach to mitigate overhead of sqs queue deployments, would be to attach the name of the table to the queued item, deqeue, check if it belongs to a paused table, if it doesn't then process, otherwise put it back on the queue. Further clarification would be needed to address potential concerns of this approach.

Example flow:
```shell
hacked scratchdata % curl "http://localhost:8420/api/data/query?api_key=local&query=select%20*%20from%20events"                        

Catalog Error: Table with name events does not exist!
Did you mean "pg_views"?
hacked scratchdata % curl -X POST "http://localhost:8420/api/data/insert/events?api_key=local" --json '{"user": "alice", "event": "touch"}'
ok%                                                                                                                                                hacked scratchdata % curl "http://localhost:8420/api/data/query?api_key=local&query=select%20*%20from%20events"                            

[
        {"__row_id":1771673417511530496,"user":"alice","event":"touch"}
]
hacked scratchdata % curl -X POST "http://localhost:8420/api/data/pause/events?api_key=local&paused=true"                                 
ok%                                                                                                                                                
hacked scratchdata % curl -X POST "http://localhost:8420/api/data/insert/events?api_key=local" --json '{"user": "bob", "event": "touch"}'
ok%                                                                                                                                                hacked scratchdata % curl "http://localhost:8420/api/data/query?api_key=local&query=select%20*%20from%20events"           

[
        {"__row_id":1771673417511530496,"user":"alice","event":"touch"}
]
hacked scratchdata % curl -X POST "http://localhost:8420/api/data/pause/events?api_key=local&paused=false"                                 
ok%                                                                                                                                                
hacked scratchdata % curl "http://localhost:8420/api/data/query?api_key=local&query=select%20*%20from%20events"           

[
        {"__row_id":1771673417511530496,"user":"alice","event":"touch"},
        {"__row_id":1771673478022754304,"user":"bob","event":"touch"}
]
```


pkg/workers/workers.go
```diff
@@ -22,32 +20,29 @@ type ScratchDataWorker struct {
...
 
 	for {
-		item, ok := w.StorageServices.Queue.Dequeue()
-
-		if !ok {
-			time.Sleep(1 * time.Second)
-		} else {
-			message, err := w.messageToStruct(item)
+		select {
+		case <-ctx.Done():
```
poundifdef, line 28: I actually intentionally don't use a channel to dequeue messages, and here is why. (tell me if there is a logic error.) Suppose this sequence of events:

1. queue.Dequeue() fetches a message from SQS
2. We attempt to send that message to the channel, but it blocks because the worker is busy
3. We ctrl-c the program

In this case, I think we'd lose a message. ctx.Done() will trigger, this worker will return and the program will terminate. Meanwhile we've already fetched a message from SQS, but it hasn't been consumed by anyone. 

(No need to fix this now, just something to talk through!)

poundifdef:
Quick thought: I wonder if we could use dead-letter queues to implement a pause functionality by toggling where new messages are sent.

pkg/workers/workers.go
breadchris: yes, this is a failure scenario. this is actually something I noticed in the code. The message should only be deleted from the queue if it is handled successfully https://chat.openai.com/share/1b1b2c58-6ee5-4b3f-b0f3-67260eb71780

The queue delete needs to be moved out of this function and is called after a successful processing https://github.com/scratchdata/scratchdata/blob/main/pkg/storage/queue/sqs/sqs.go#L75

breadchris:
If you reject a message every time it belongs to a paused table, it will go to the back of the queue. A DLQ would be helpful if we want to keep track of paused queues that are putting pressure on the queue.

pkg/workers/workers.go
poundifdef: That seems like a lot of work just so we can a channel, doesn't it? To me, the question is whether we want consumers from the queue to have to contend with a `Delete()` operation in addition to Dequeue?

Other things to think about: What timeout should we set before SQS automatically re-queues an unacknowledged message? If we end up using Redis as a queue in the future (ie, `BLPOP`) then consuming+deleting are the same action.

poundifdef:
> If you reject a message every time it belongs to a paused table, it will go to the back of the queue. A DLQ would be helpful if we want to keep track of paused queues that are putting pressure on the queue.

I worry about chattiness+starvation here: re-queuing items over and over again, and consuming workers with the busywork of resending paused items. 

I'm wondering if SQS is the right queue to use after all.

breadchris:
This is another YC company (also go OSS): https://hatchet.run/ they use rabbitmq currently for their queue.

breadchris:
from our conversation, we concluded that existing throughput can be handled by an SQL queue until we "feel the pain". When we hit that point, redis could be an option to explore.

PR #135, opened 2024-03-24
AWS Secret Manager for DB credentials 

### LINKED ISSUE #126: Use a vault for DB credentials

mehulmathur16:
/claim #126 
/fixes #126 

Hi @poundifdef, I have tested out the memory vault & aws vault & it is working as expected. 

The aws config looks like this :

```
vault:
  type: aws
  settings:
    AccessKey: "YOUR_AWS_ACCESS_KEY"
    SecretKey: "YOUR_AWS_SECRET_KEY"
    Prefix: "YOUR_PREFIX"
    Region: "us-east-1"
```

### DIFF
--- a/pkg/destinations/destinations.go
+++ b/pkg/destinations/destinations.go
@@ -2,8 +2,10 @@ package destinations
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"io"
+	"strconv"
 
 	"github.com/EagleChen/mapmutex"
 	"github.com/rs/zerolog/log"
@@ -14,6 +16,14 @@ import (
 	"github.com/scratchdata/scratchdata/pkg/destinations/duckdb"
 	"github.com/scratchdata/scratchdata/pkg/destinations/redshift"
 )
+
+type Creds struct {
+	ID       int64          `yaml:"id" json:"id"`
+	Type     string         `yaml:"type" json:"type"`
+	Name     string         `yaml:"name" json:"name"`
+	Settings map[string]any `yaml:"settings" json:"settings"`
+	APIKeys  []string       `yaml:"api_keys" json:"api_keys"`
+}
 
 type DestinationManager struct {
 	storage *models.StorageServices
@@ -83,16 +93,20 @@ func (m *DestinationManager) Destination(ctx context.Context, databaseID int64) 
 		defer m.mux.Unlock(databaseID)
 
 		var dest Destination
+		var creds Creds
 
 		dest, ok := m.pool[databaseID]
 		if ok {
 			return dest, nil
 		}
 
-		creds, err := m.storage.Database.GetDestinationCredentials(ctx, databaseID)
+		jsonCreds, err := m.storage.Vault.GetCredential(strconv.Itoa(int(databaseID)))
 		if err != nil {
 			return nil, err
 		}
+
+		credsBytes := []byte(jsonCreds)
+		err = json.Unmarshal(credsBytes, &creds)
 
 		switch creds.Type {
 		case "duckdb":
--- a/pkg/storage/vault/aws/aws.go
+++ b/pkg/storage/vault/aws/aws.go
@@ -1,0 +1,94 @@
+package aws
+
+import (
+	"context"
+	"errors"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	awsConfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/credentials"
+	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
+)
+
+type AWSVault struct {
+	client *secretsmanager.Client
+	prefix string
+}
+
+func NewAWSVault(conf map[string]any) (*AWSVault, error) {
+	cfg, err := awsConfig.LoadDefaultConfig(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	var accessKey, secretKey, prefix, region string
+
+	if accessKeyVal, ok := conf["AccessKey"].(string); ok {
+		accessKey = accessKeyVal
+	} else {
+		return nil, errors.New("AccessKey not found or not a string")
+	}
+
+	if secretKeyVal, ok := conf["SecretKey"].(string); ok {
+		secretKey = secretKeyVal
+	} else {
+		return nil, errors.New("SecretKey not found or not a string")
+	}
+
+	if prefixVal, ok := conf["Prefix"].(string); ok {
+		prefix = prefixVal
+	} else {
+		return nil, errors.New("prefix not found or not a string")
+	}
+
+	if regionVal, ok := conf["Region"].(string); ok {
+		region = regionVal
+	} else {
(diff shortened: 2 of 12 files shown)


mehulmathur16:
The below video shows the working with both vaults.


https://github.com/scratchdata/scratchdata/assets/64700961/90f70c22-e7e6-43f7-b1bd-f53900e43391

On AWS console:

image

image

mehulmathur16:
Hi @poundifdef, can you please review this PR ?

pkg/storage/vault/aws/aws.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/secretsmanager"
+	"github.com/scratchdata/scratchdata/config"
+)
+
+type AWSVault struct {
+	client *secretsmanager.Client
+	prefix string
+}
+
+func NewAWSVault(conf map[string]any, destinations []config.Destination) (*AWSVault, error) {
```
poundifdef, line 21: Do not pass `[]config.Destination` into this function. That should only be passed to the in-memory implementation.

pkg/storage/vault/aws/aws.go
```diff
@@ -0,0 +1,135 @@
...
+	cfg, err := awsConfig.LoadDefaultConfig(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	var accessKey, secretKey, prefix, region string
+
+	if accessKeyVal, ok := conf["AccessKey"].(string); ok {
```
poundifdef, line 29: All of these configurations should be part of the `AWSVault` struct.

pkg/storage/vault/aws/aws.go
```diff
@@ -0,0 +1,135 @@
...
+
+	cfg.Region = region
+	client := secretsmanager.NewFromConfig(cfg)
+
+	vault := &AWSVault{
+		client: client,
+		prefix: prefix,
+	}
+
+	for _, dest := range destinations {
```
poundifdef, line 65: There is no need to upload destinations to AWS. Either credentials will be read from a config file, or they will be read from AWS, depending on how the vault is configured.

pkg/storage/vault/vault.go
```diff
@@ -0,0 +1,23 @@
...
+	GetCredential(name string) (string, error)
+	SetCredential(name, value string) error
+}
+
+func NewVault(vaultConf config.Vault, destinations []config.Destination) (Vault, error) {
+	switch vaultConf.Type {
+	case "memory":
+		return memory.NewMemoryVault(destinations)
+	case "aws":
+		return aws.NewAWSVault(vaultConf.Settings, destinations)
```
poundifdef, line 19: Do not pass `destinations` to this function.

pkg/destinations/destinations.go
```diff
@@ -83,17 +93,21 @@ func (m *DestinationManager) Destination(ctx context.Context, databaseID int64)
...
 		var dest Destination
+		var creds Creds
 
 		dest, ok := m.pool[databaseID]
 		if ok {
 			return dest, nil
 		}
 
-		creds, err := m.storage.Database.GetDestinationCredentials(ctx, databaseID)
+		jsonCreds, err := m.storage.Vault.GetCredential(strconv.Itoa(int(databaseID)))
```
poundifdef, line 103: We only need to store the value of `Settings` in the vault. Everything else we should fetch from the database. Perhaps the [GetDestinationCredentials()](https://github.com/scratchdata/scratchdata/blob/main/pkg/storage/database/database.go#L17) function should be renamed to `GetDestination()` and return all the same values except for Settings.

poundifdef (changes requested):

pkg/storage/vault/vault.go
mehulmathur16: @poundifdef we need to pass it to the memory case, right ?

pkg/destinations/destinations.go
mehulmathur16: @poundifdef I need some more clarity on this point. Currently the memory vault has this param : `destinations []config.Destination`. Do you want me to remove Settings from Destination struct ?

If yes, then what should be the new param to the function ?

mehulmathur16:
Hi @poundifdef, I have resolved the above comments. Can you please review ?

mehulmathur16:
Hi @poundifdef , can you please review the updated changes ?

poundifdef:
I haven't forgotten, I'll be sure to let you know when I've tested. Thanks!

mehulmathur16:
Sure @poundifdef 👍

mehulmathur16:
Hi @poundifdef, There appear to be a lot of structural changes on main branch. My work was based on the old structure as described by you in the issue description. Can you tell if the requirements are a bit changed now ?

mehulmathur16:
@poundifdef can you update on the above ?
I don't think it will take much time to test this one. I've invested quite a good amount of time on this issue.

pkg/storage/vault/aws/aws.go
poundifdef: Please use the exact same pattern as you used [here](https://github.com/scratchdata/scratchdata/pull/120/files#diff-39009ce4deb7c44edd1c180471d6d79ea36e471a012dd4725443057dbdb94870R14) to automatically map configuration parameters from the yaml file into the struct.

pkg/storage/vault/memory/memory.go
```diff
@@ -0,0 +1,42 @@
...
+type MemoryVault struct {
+	destinations map[string]map[string]any
+}
+
+func NewMemoryVault(destinations []config.Destination) (*MemoryVault, error) {
+	vault := &MemoryVault{
+		destinations: make(map[string]map[string]any),
+	}
+	for _, dest := range destinations {
+		vault.destinations[strconv.Itoa(int(dest.ID))] = dest.Settings
```
poundifdef, line 21: If there are multiple destinations, then this code does not work. Try it and you'll see.

pkg/storage/vault/aws/aws.go
poundifdef: Please use the same names of the aws yaml configuration key as we're using elsewhere in the code. https://github.com/scratchdata/scratchdata/blob/main/pkg/storage/queue/sqs/sqs.go#L17-L19

poundifdef (changes requested):
You are correct that the project structure has changed on the `main` branch, however, it is okay for you to keep the current structure of the code. I will handle rebasing myself.

mehulmathur16:
@poundifdef I have resolved your comments. Please have a look.

pkg/storage/vault/memory/memory.go
mehulmathur16: I think this was happening because I was fetching credentials using databaseId earlier, while it should be destination credential ID.

mehulmathur16: This has been resolved now. We just need to make sure that we are using different destination IDs in config yaml.

mehulmathur16:
@poundifdef are we good to go with the new changes ?

poundifdef:
Looks good. Thank you.

PR #139, opened 2024-03-26
Move frontend into repo

### LINKED ISSUE #136: move frontend into repo
Currently, the front end is in another repo and should be moved over to this one. 

There are currently multiple versions of the frontend, svelte and HTML templates. Go HTML templates with tailwindcss have been chosen as the tech to use moving forward.

Questions:
- Should the svelte code also come over for reference, or should it be kept in the other repo? Answer seems like it should be kept where it is and only relevant code comes over.

Screenshot 2024-03-25 at 1 01 30 PM

breadchris:
Closes #136


pkg/api/router.go
```diff
@@ -55,5 +43,40 @@ func CreateMux(apiFunctions ScratchDataAPI) *chi.Mux {
...
+	d, err := view.GetView()
+	if err != nil {
+		panic(err)
+	}
+	router.Mount("/dashboard", d)
+
+	fileServer := http.FileServer(http.FS(static.Static))
+	router.Handle("/static/*", http.StripPrefix("/static", fileServer))
+
+	r.Mount("/", router)
```
poundifdef, line 80: Check if if `DashboardConfig.Enabled` is true.

pkg/storage/database/database.go
```diff
@@ -18,19 +15,13 @@ type Database interface {
...
+	GetAPIKeyDetails(ctx context.Context, hashedAPIKey string) (APIKey, error)
 
 	CreateShareQuery(ctx context.Context, destId int64, query string, expires time.Duration) (queryId uuid.UUID, err error)
-	GetShareQuery(ctx context.Context, queryId uuid.UUID) (models.SharedQuery, bool)
+	GetShareQuery(ctx context.Context, queryId uuid.UUID) (SharedQuery, bool)
 
-	Hash(s string) string
-}
-
-func NewDatabaseConnection(conf config.Database, destinations []config.Destination, adminKeys []config.APIKey) Database {
```
poundifdef, line 29: This is the behavior I want:

- If the user is configuring Scratch via config file, then we should only read values from that config. (Destinations, API Keys, etc.) We could even name this database "config" or "static". In this case, the user _cannot_ add a new destination. This is so there is a very clear single source of truth for where data is coming from - either a database or a config file. 
- Otherwise, read _all_ values from the database, and ignore any destinations or API keys defined in the config.
- Let's decouple Gorm. We can have a common Gorm implementation or set of utilities but inevitably sqlite and postgres will have differences.
- In the future, I might want to add dynamodb or firestore so that this can be deployed as a completely serverless solution. So decoupling from Gorm is the way to go.

That means

pkg/view/manager.go
```diff
@@ -0,0 +1,132 @@
...
+		tfs:       fsys,
+		watcher:   watcher,
+	}
+	err = m.loadTemplates()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to load templates")
+	}
+
+	// TODO breadchris setup fs different in dev for live reload
+	//go m.watchTemplateChanges()
```
poundifdef, line 45: i'm fine with a config option that reads templates from a local filesystem for local dev and uses go:embed for production.

pkg/api/router.go
```diff
@@ -55,5 +43,40 @@ func CreateMux(apiFunctions ScratchDataAPI) *chi.Mux {
 
 	r.Mount("/api", api)
 
+	router := chi.NewRouter()
+	router.Use(middleware.Logger)
+	router.Use(jwtauth.Verifier(apiFunctions.tokenAuth))
+
+	// TODO breadchris renable auth
+	//router.Use(apiFunctions.Authenticator(apiFunctions.tokenAuth))
```
poundifdef, line 51: You've glossed over something pretty tricky. Maybe there's a way for this to be pluggable for a proprietery version.

- If a user is running this for the first time (the `go run` case), we shouldn't have auth. They should just be able to use the thing.
- The common case for a deployment would be for everyone to be part of the same organization. So any login (whether basic auth or google sign in) would assume that anyone has access to the same resources. one-to-many for teams and users.
- The proprietary case is for there to be multiple users and multiple teams. Each user only has access to the resources for the teams they belong to. many to many for teams and users. This is only relevant in a hosted version of Scratch Data.

pkg/view/manager.go
```diff
@@ -0,0 +1,132 @@
...
+		handler: nil,
+		tmpl:    tmpl,
+	}
+
+	// TODO breadchris need setup fs different in dev for live reload
+	//return s.watcher.Add(path)
+	return nil
+}
+
+func (s *Manager) BuildRouter() (*chi.Mux, error) {
```
poundifdef, line 118: I'm not totally sure how any of this works as there is a lot of autowiring happening.

I want there to be a function that looks like this: https://github.com/go-gitea/gitea/blob/main/routers/web/web.go#L405

Where do we define the logic for a route?

poundifdef:
Unanswered question is where should business logic live?

We want a user to be able to add a new destination both via an API and the GUI. Right now all of that logic lives in the http handler. I think we probably need a "controllers" abstraction for this case.

breadchris:
> Unanswered question is where should business logic live?
> 
> We want a user to be able to add a new destination both via an API and the GUI. Right now all of that logic lives in the http handler. I think we probably need a "controllers" abstraction for this case.

I like to think of "let the abstraction show itself to you." If our current goal is to let people connect data sources, the most important thing is getting that shipped. Since we are only a couple of people at the moment, it is easier to spot and refactor into abstractions.

pkg/view/manager.go
breadchris: I knew you were going to pick up on this one ;) I was coding this one a little more for fun than what I think should be in here right now. I agree that we should have it look like gitea's for simplicity.

We do need logic setup so that we can load templates from disk during development and then from the embed in prod (ex. config option). This will help speed up local dev significantly as UI changes are immediate.

breadchris: I was playing with the feeling of being able to just drop HTML files in this directory and having something happen. 

Screenshot 2024-03-26 at 9 48 52 AM

poundifdef (approved):
Let me know when you've rebased and I'll merge.

Issue #138, opened 2024-03-25
duckdb integration


breadchris:
As a user with a duckdb instance, I should be able to log into the dashboard, and configure a duckdb destination for data based on my config. This includes a token and a database name.

PR #145, opened 2024-03-28
new, edit, delete duckdb

breadchris:
#138 

new
Screenshot 2024-03-28 at 2 52 42 PM

edit/delete
Screenshot 2024-03-28 at 2 52 49 PM

Other providers have been greyed out until we have support for them:
Screenshot 2024-03-28 at 2 52 35 PM


poundifdef:
Awesome! I played with this locally. Some high-level comments about the user experience:

- **API Keys.** When the user adds a connection, let's create and show them an API key immediately. That way they can get started right away. (Bonus: show them a snippet they can paste. In the future this can have many languages.)
- **Walidate connection.** We should test the connection before saving. That was the intention of this [TestCredentials()](https://github.com/scratchdata/scratchdata/blob/main/pkg/destinations/destinations.go#L61) function and an initial (unused) implementation of adding new destinations [here](https://github.com/scratchdata/scratchdata/blob/495608968d2e5126030b1d567416d95547d08bb5/pkg/api/destinations.go#L37).
- **List connections screen.** The "Connect Database" block is only supposed to show for the empty state when there are no connections. If there _are_ connections, it should look like this: 
Screen Shot 2024-03-28 at 9 17 37 PM
- We probably should not show the password on the screen when a user is editing a credential. If we need a sensitive flag around these fields we can add that.

pkg/storage/database/gorm/gorm.go
```diff
@@ -139,49 +139,50 @@ func (s *Gorm) getTeamId(userId uint) uint {
...
-// AddAPIKey implements database.ProprietaryDB.
 func (*Gorm) AddAPIKey(ctx context.Context, destId int64, hashedAPIKey string) error {
 	panic("unimplemented")
 }
 
-// CreateDestination implements database.ProprietaryDB.
-func (s *Gorm) CreateDestination(ctx context.Context, userId uint, destType string, settings map[string]any) (config.Destination, error) {
+func (s *Gorm) CreateDestination(
+	ctx context.Context,
+	userId uint,
```
poundifdef, line 148: This should really be a `teamId`. I feel like the userId -> teamId translation should happen before this function is called.

pkg/view/csrf.go
```diff
@@ -0,0 +1,69 @@
+package view
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/base64"
+	"net/http"
+)
+
+func generateCSRFToken() (string, error) {
```
poundifdef, line 10: Is there a technical reason not to use this: https://github.com/gorilla/csrf/tree/main

You're already importing it in `go.mod`. If there is a specific technical deficiency with that library then I'm open to hearing but otherwise there's no need for us to maintain our own implementation of something which is not core to the product.

pkg/view/router.go
```diff
@@ -54,17 +90,159 @@ func New(c config.DashboardConfig, auth func(h http.Handler) http.Handler) (*chi
...
 
 	r.Get("/connections/new", func(w http.ResponseWriter, r *http.Request) {
-		err := gv.Render(w, http.StatusOK, "pages/connections/new", loadModel(r))
+		err := gv.Render(w, http.StatusOK, "pages/connections/upsert", loadModel(r))
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+		}
+	})
+
+	r.Post("/connections/new", func(w http.ResponseWriter, r *http.Request) {
```
poundifdef, line 120: In an ideal world, the logic here would be decoupled from this route. The next feature (next week's demo) is for users give plaid-like links to their customers to fill in their own connection details. I'd love to be able to re-use the same logic without much ado.

poundifdef:
Instead of graying out unsupported connections, let's just omit them from the page.

pkg/view/csrf.go
breadchris: I swear I ran into [this bug](https://github.com/gobuffalo/buffalo/issues/1067), which I have used this library before. But now that I am putting together a minimal reproduction case, I am not getting the bug to trigger. 

The problem I have seen is that the library requires `SESSION_SECRET` set in the environment, which is an undocumented environmental dependency. This design pattern leads to many head-scratching problems both in local dev and production, and I avoid code that has it.

I am also perusing the libraries' source code to try to find the source of the bug, but my static analysis eyes can't find it.

Let me try the library again and see if I get that problem...

pkg/view/router.go
breadchris: The flow of development that I generally practice for things like this is breaking code shared code into functions. Dependencies declared in the parameters, and no side effects, ie. functional code.

This makes the function very portable and can be easily refactored to anywhere else in the code. As multiple of these functions seem to want to be grouped together (similar dependencies), they get their own service (type Service struct {}).

idk if this makes sense, but the goal is to optimize for velocity while also balancing abstraction needs.

pkg/view/csrf.go
breadchris: alright, odd, I wasn't able to repro. I agree it should be gorilla/csrf, but be warned that I ran into that bug again somehow.

poundifdef:
Would you mind updating goland to use `gooimports` sorting for this project? Getting a lot of extra diffs because of import statements moving things around. https://youtrack.jetbrains.com/issue/GO-10293/Set-goimports-as-a-default-import-sorting#focus=Comments-27-4597024.0-0

pkg/view/router.go
```diff
@@ -1,49 +1,104 @@
...
-func New(c config.DashboardConfig, auth func(h http.Handler) http.Handler) (*chi.Mux, error) {
+func New(
+	storageServices *storage.Services,
+	c config.DashboardConfig,
+	destManager *destinations.DestinationManager,
+	auth func(h http.Handler) http.Handler,
+) (*chi.Mux, error) {
 	r := chi.NewRouter()
 
+	csrfMiddleware := csrf.Protect([]byte(c.CSRFSecret))
```
poundifdef, line 55: the `/logout` route should just clear this too. Maybe it should just iterate over every cookie and clear it.

pkg/view/router.go
```diff
@@ -54,17 +109,220 @@ func New(c config.DashboardConfig, auth func(h http.Handler) http.Handler) (*chi
...
+		}
+
+		cd := config.Destination{
+			Type:     t,
+			Settings: settings,
+		}
+
+		err = destManager.TestCredentials(cd)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
```
poundifdef, line 193: It would be way better to flash an error message, and then pre-populate the form with whatever fields they entered. I think gorilla will do message flashing: https://github.com/gorilla/sessions

Here's the tailwind code for an alert:

```
<div class="rounded-md bg-red-50 p-4">
  <div class="flex">
    <div class="flex-shrink-0">
      <svg class="h-5 w-5 text-red-400" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true">
        <path fill-rule="evenodd" d="M10 18a8 8 0 100-16 8 8 0 000 16zM8.28 7.22a.75.75 0 00-1.06 1.06L8.94 10l-1.72 1.72a.75.75 0 101.06 1.06L10 11.06l1.72 1.72a.75.75 0 101.06-1.06L11.06 10l1.72-1.72a.75.75 0 00-1.06-1.06L10 8.94 8.28 7.22z" clip-rule="evenodd" />
      </svg>
    </div>
    <div class="ml-3">
      <h3 class="text-sm font-medium text-red-800">There were 2 errors with your submission</h3>
      <div class="mt-2 text-sm text-red-700">
        <ul role="list" class="list-disc space-y-1 pl-5">
          <li>Your password must be at least 8 characters</li>
          <li>Your password must include at least one pro wrestling finishing move</li>
        </ul>
      </div>
    </div>
  </div>
</div>
```

pkg/view/router.go
```diff
@@ -54,17 +109,220 @@ func New(c config.DashboardConfig, auth func(h http.Handler) http.Handler) (*chi
...
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		m := loadModel(r)
+		m.Connect.InsertExample = fmt.Sprintf(
+			"curl -X POST '%s/api/data/insert/%s?api_key=%s' --json '{\"user\": \"bob\", \"event\": \"click\"}'",
+			c.ExternalURL,
+			dest.Name,
```
poundifdef, line 221: This is confusing. What if the name has spaces? I'd just use an obvious value like `your_table`.

pkg/view/templates/layout/base.html
```diff
@@ -61,33 +61,6 @@
                     <span class="ms-3">API Keys</span>
```
poundifdef, line 61: Any reason you're leaving this link here? (The page doesn't do anything.)

breadchris: nope, gone

pkg/view/router.go
breadchris: if the name of the table is `your_table` then the copy and paste wouldn't work until they edit in their own table name. Is that what you had in mind?

poundifdef: I'd say we give a default value, like `scratchdata_test` or `scratchdata` or even `your_table`. The software will automatically create the table if it doesn't exist when they do the insert, and perhaps that will help users see the end-to-end flow.

breadchris: ah ok, I think I had a misunderstanding of what a table was. This makes sense.

pkg/view/router.go
```diff
@@ -1,74 +1,386 @@
...
+
+		cd := config.Destination{
+			Type:     t,
+			Settings: settings,
+		}
+
+		err = destManager.TestCredentials(cd)
+		if err != nil {
+			log.Err(err).Msg("failed to connect to destination")
+			newFlash(w, r, FlashTypeError, "Failed to connect to destination. Check the settings and try again.")
```
poundifdef, line 256: I'm fine with returning the error string from line 253 since it will have a more descriptive error.

pkg/view/router.go
```diff
@@ -1,74 +1,386 @@
...
+			http.Redirect(w, r, "/dashboard/connections/new/"+t, http.StatusFound)
```
poundifdef, line 257: If there is an error, let's preserve the fields the user already entered. Right now they don't get an indication of what went wrong and have to re-enter everything from scratch.

pkg/view/router.go
```diff
@@ -1,74 +1,386 @@
...
+			return
+		}
+
+		settings := map[string]any{}
+		name := r.Form.Get("name")
+
+		t := r.Form.Get("type")
+		switch t {
+		case "duckdb":
+			// XXX breadchris support file destination?
```
poundifdef, line 226: Ignore the file destination for now. We never want a file destination for the hosted version, but self-hosted users might want this. Not sure how to differentiate these paths.

pkg/storage/database/database.go
```diff
@@ -26,6 +28,7 @@ type Database interface {
 	GetShareQuery(ctx context.Context, queryId uuid.UUID) (models.SharedQuery, bool)
 
 	GetUser(int64) *models.User
+	GetTeamId(userId uint) (uint, error)
```
poundifdef, line 31: This is a nitpick we can change in a future PR: 

With our data model, a user can belong to multiple teams. So if you call `GetUser()`, then `models.User` also return a `[]Team`.

I'd suggest having the router be responsible for figuring out the team related to our request. Right now a user can only belong to one team, but in the future we'll add a level of indirection: `/dashboard/<team>/...` so we know how to manage requests.

Not a change that _needs_ to be made right now but since this is new code I wanted to point it out.

poundifdef:
Awesome, this is looking really good. Thank you!

PR #148, opened 2024-04-02
use sqlite for static db

poundifdef:


pkg/storage/database/gorm/gorm.go
```diff
@@ -65,11 +65,9 @@ func NewGorm(
 
 	var teamCount int64
 	db.Model(&models.Team{}).Count(&teamCount)
-	if teamCount == 0 {
-		if rc.DefaultUser == "" {
-			return nil, errors.New("Must specify a default_user in the DB settings file")
-		}
 
+	// Bootstrap a default user if we specify one
```
breadchris, line 69: I try to keep database methods less opinionated in how they operate, and defer to the caller on how this type of logic should be handled. It makes database read/writes more generally applicable.

I would have this method accept a "userId" parameter to control who the default user is.

poundifdef: Removed this - there's no need for a default user since users are created when logged into the dashboard.

Issue #147, opened 2024-04-02
credential share link


breadchris:
notes on this feature:
```
- share link
	- plaid experience
	- receive a share link
	- option to choose destination
	- big query
	- enter big query creds
	- save
	- congrats/error on save
	- paypal, if customer A specifies redirect, go to another page
	- webhook after connect
	- your connections, external connections?
	- maybe pending status
	- concerns
		- expiration
		- know what share links are out there
```

PR #150, opened 2024-04-02
Remote Connection URL

breadchris:
A user should be able to create a URL to share with someone they wish to send data to.

For example, Customer A uses scratchdata and Customer B is a customer of Customer A. Customer B has a Big Query instance and would like to get data from Customer A pushed to them. Customer A would create a Remote Connection URL and share it with Customer B. Customer B enters in their connection information for Big Query, saves it, and it is now accessible to Customer A to start sending data to it.

#147


poundifdef:
Overall - I am able to get it to work! Yay! Some comments:

- It seems like these links are reusable. It doesn't look like `DeleteConnectionRequest()` is used anywhere. Also seems like you can resubmit the form if you hit "refresh" on the success page.
- If you enter a bad credential, the page correctly shows an error, however it also shows the dashboard side bar. So there must be some code path we're using the wrong template.

In an ideal world, the end user would be able to also choose their database type, instead of a link being specific to duckdb. However, we can table that for now.

I'm fine with merging this and then making tickets for the above issues and tackling them on another iteration, since i know you've already started on the other feature.

pkg/view/router.go
```diff
@@ -471,11 +491,12 @@ func New(
...
 
-			m.HideSidebar = true
-			err = gv.Render(w, http.StatusOK, "pages/request/success", m)
+			err = storageServices.Database.DeleteConnectionRequest(r.Context(), connReq.ID)
 			if err != nil {
-				http.Error(w, err.Error(), http.StatusInternalServerError)
+				log.Err(err).Msg("failed to delete connection request")
 			}
+
+			http.Redirect(w, r, "/dashboard/request/success", http.StatusFound)
```
poundifdef, line 499: There is an edge case where the user hits the back button, then they end up on this page. However it correctly throws an error saying "connection request not found" so i think we can ignore that for now.

PR #157, opened 2024-04-10
shared query page

breadchris:
Screenshot 2024-04-10 at 1 05 15 PM

After creating a shared query
```
curl -X POST "http://localhost:8080/api/data/query/share?api_key=xxx" --data '{"query":"select * from events","duration":10000,"name":"Shared data"}'
```

You can link an unauthenticated user a link to download the data:
http://localhost:8080/share/<shared query id>

### DIFF
--- a/pkg/storage/database/models/models.go
+++ b/pkg/storage/database/models/models.go
@@ -8,17 +8,11 @@ import (
 	"gorm.io/gorm"
 )
 
-type SharedQuery struct {
-	ID            string
-	Query         string
-	DestinationID int64
-	ExpiresAt     time.Time
-}
-
-type ShareLink struct {
+type ShareQuery struct {
 	gorm.Model
-	UUID          string `gorm:"index:idx_uuid,unique"`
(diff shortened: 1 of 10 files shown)


pkg/storage/database/models/models.go
```diff
@@ -8,17 +8,11 @@ import (
 	"gorm.io/gorm"
 )
 
-type SharedQuery struct {
```
breadchris, line 11: It seemed like SharedQuery and ShareLink were duplicates, I kept ShareQuery.

pkg/view/share.go
```diff
@@ -0,0 +1,88 @@
...
+
+		queryUUID := chi.URLParam(r, "uuid")
+
+		id, err := uuid.Parse(queryUUID)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+
+		cachedQuery, found := storageServices.Database.GetShareQuery(r.Context(), id)
```
breadchris, line 63: This code is copied from where it is used in API. As a part of the refactor, this functionality should be easier to reuse.

PR #160, opened 2024-04-15
api key cache ttl

### LINKED ISSUE #159: Cache API keys
For the API, we currently fetch API key details on every request. This check happens [here](https://github.com/scratchdata/scratchdata/blob/main/pkg/api/auth.go#L41).

This is a "hot path" and we should cache API key results for some configurable amount of time (ie, 30s.) While we're at it, it will _probably_ be better to add the API key details to the context rather than just specific fields related to the key. I have an example [here](https://github.com/scratchdata/scratchdata/blob/query-as-api/pkg/api/auth.go#L51).

Finally, there is a bug where we're not correctly fetching team data from gorm. [This](https://github.com/scratchdata/scratchdata/blob/mongodb-source/pkg/storage/database/gorm/gorm.go#L313) commit fixes that.

breadchris:
resolves #159

### DIFF
--- a/pkg/api/api.go
+++ b/pkg/api/api.go
@@ -5,14 +5,17 @@ import (
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
+	"net/http"
+	"time"
+
 	"github.com/go-chi/jwtauth/v5"
+	"github.com/jellydator/ttlcache/v3"
 	"github.com/scratchdata/scratchdata/pkg/config"
 	"github.com/scratchdata/scratchdata/pkg/storage"
+	"github.com/scratchdata/scratchdata/pkg/storage/database/models"
 	"github.com/scratchdata/scratchdata/pkg/util"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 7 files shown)


pkg/api/api.go
```diff
@@ -52,6 +56,15 @@ func NewScratchDataAPI(
 		return nil, err
 	}
 
+	apiKeyTTL := conf.API.APIKeyCacheTTL
+	if apiKeyTTL == 0 {
```
poundifdef, line 60: Does a TTL of 0 mean "we do not cache anything"? If so, then i'm okay removing this if-statement so that there's some way to disable the cache.

breadchris: For this library, the answer seems to be yes and no? While the item will instantly expire, any lookup that is less than an hour https://github.com/jellydator/ttlcache/blob/v3/cache.go#L537 will succeed, but return an item that is expired, which you could check with `item.ExpiresAt() < time.Now()`.

PR #161, opened 2024-04-15
db credential vault

### LINKED ISSUE #126: Use a vault for DB credentials

breadchris:
Closes #126


pkg/destinations/destinations.go
```diff
@@ -83,24 +86,71 @@ func (m *DestinationManager) TestCredentials(creds config.Destination) error {
...
+		)
 
 		dest, ok := m.pool[databaseID]
 		if ok {
 			return dest, nil
 		}
 
+		// XXX breadchris does the caller verify existence and authz for this database?
+		// is this call needed?
 		creds, err := m.storage.Database.GetDestinationCredentials(ctx, databaseID)
```
poundifdef, line 140: I'd suggest deleting this from the `database` interface completely. If you fix the resulting compiler errors then everything will probably Just Work.

breadchris: This change will be a bit involved since there are multiple callers of `DestinationManager.Destination` where they do not have the database `type` in scope. 

Perhaps a follow up to this one?

poundifdef: Now that I've looked in more detail I see how it works. Probably this function should just be renamed `GetDestination`. And yeah, I think we do want to keep this function to make sure the destination actually exists and is valid before trying to fetch credentials from a vault.

pkg/destinations/destinations.go
```diff
@@ -83,24 +87,76 @@ func (m *DestinationManager) TestCredentials(creds config.Destination) error {
...
 		if err != nil {
 			return nil, err
 		}
 
-		settings := creds.Settings.Data()
+		cn := credentialName(creds.TeamID, databaseID, creds.Name)
```
poundifdef, line 150: Shall we store the credential name in the database in the Destination model? This way when we inevitably change the naming scheme we won't lose anything. Also useful to track down orphaned items in the vault.

pkg/storage/vault/aws/aws.go
```diff
@@ -0,0 +1,104 @@
...
+
+	if vault.AccessKeyId == "" {
+		return nil, errors.New("AccessKeyId not found or not a string")
+	}
+
+	if vault.SecretAccessKey == "" {
+		return nil, errors.New("SecretAccessKey not found or not a string")
+	}
+
+	if vault.Prefix == "" {
```
poundifdef, line 39: A prefix isn't required if the user doesn't want to specify one. (I know this was copied from the previous PR.)

pkg/storage/vault/aws/aws.go
```diff
@@ -0,0 +1,104 @@
...
+	cfg.Credentials = credentials.NewStaticCredentialsProvider(vault.AccessKeyId, vault.SecretAccessKey, "")
+	cfg.Region = vault.Region
+	client := secretsmanager.NewFromConfig(cfg)
+	vault.Client = client
+
+	return &vault, nil
+}
+
+func (v *AWSVault) GetCredential(name string) (string, error) {
+	secretName := fmt.Sprintf("%s-%s", v.Prefix, name)
```
poundifdef, line 61:
```suggestion
	secretName := fmt.Sprintf("%s%s", v.Prefix, name)
```

A prefix might be `scratchdata/production/`

poundifdef:
`service.go:DeleteDestination()` should also remove from vault.

PR #162, opened 2024-04-17
Updates Dockerfile to fix docker build, adds docker-compose file

NortySpock:
Please let me know if this solution is unworkable for some reason, or if you need a CLA or something from me.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
-FROM golang:1.21.1 as builder
+FROM golang:1.22.1 as builder
 WORKDIR /build
 COPY go.mod go.sum ./
+COPY *.go ./
 RUN go mod download
-COPY ./ ./
-RUN go build scratchdata
-EXPOSE 3000
-ENTRYPOINT ["./scratchdata", "local.toml"]
+RUN go build -o scratchdata
+EXPOSE 8080
+ENTRYPOINT ["./scratchdata", "config.yaml"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,33 @@
+services:
+
+  scratchdb:
+    container_name: scratchdb
+    image: scratchdb
+    build:
+     context: .
+    ports:
+      - "8080:8080"
+
+    volumes:
+        - type: bind
+          read_only: true
+          source: ./config.yaml
+          target: /config.yaml
+
+        - type: bind
+          read_only: false
+          source: ./data
+          target: /data
+
+
+    environment:
+      - PUID=1000
+      - PGID=1000
+      - TZ=America/Chicago
+
+    logging:
+      driver: "json-file"
+      options:
+        max-size: "20m"
+        max-file: "5"
+        mode: "non-blocking"


NortySpock:
I realized I should caveat this PR slightly -- I am not a golang-native developer. You should review this code critically in that light, as though I am a junior developer who doesn't even know the ways in which this PR could be wrong.

It is very possible that I am missing something obvious about which folders should be hoisted from the code repository into the Dockerfile working directory prior to the the call to `go mod download` and `go build -o scratchdata`. For instance, I did not copy in the `/models` or `/pkg` folder when I was drafting this PR yesterday, because it appeared to work. However, reading though the import section of `main.go` this evening, I realize I probably should at least copy in the `/pkg` folder, otherwise a developer working on the code later might wonder why their changes in the `/pkg/` folder are not having any effect.

NortySpock:
Still trying to make a "better" Dockerfile (or at least one that follows the principle of least surprise) on an alternate branch here: 
https://github.com/scratchdata/scratchdata/compare/main...NortySpock:scratchdata:feature/update-developer-dockerfile-to-support-using-the-developers-packages

I didn't want to cloud up this PR with noise if you like what you see here, but I figured I'd mention that I'm toying with improvements.

Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM golang:1.21.1 as builder
+FROM golang:1.22.1 as builder
 WORKDIR /build
 COPY go.mod go.sum ./
+COPY *.go ./
 RUN go mod download
-COPY ./ ./
```
breadchris, line 5: we want to keep this line because it copies all files recursively. `COPY *.go ./` will only copy the *.go files in the current directory, but we also need all the go files in `pkg/`

docker-compose.yml
```diff
@@ -0,0 +1,33 @@
+services:
```
breadchris, line 1: Since scratchdata is a single process, there isn't a need to have docker-compose.yaml at the moment. I think what would be better is documenting how to run scratchdata as a docker command

breadchris (changes requested):

Dockerfile
NortySpock: Ok, makes sense, but I have two small concerns with `COPY . .` :
 - It hoists the default config.yaml file into the build WORKDIR. With the current default start of ENTRYPOINT ["./scratchdata", "config.yaml"], that (I think) could by default point scratchdata at the default config.yaml that got baked into the image, and they would not receive the `INF main.go:31 > No config file specified, using local default values` info message to hint to them that they are running without a config file. 
   - Possibly we should change the entrypoint to not specify a config-file by default, relying on the user to add that as a (docker) runtime argument.
   - Possibly we should a .dockerignore file that ignores config.yaml so the config defaults (or sensitive information like the `crypto: jwt_private_key:` that's in there now doesn't just get baked directly into the docker image.)
 - If a user accidentally creates a data directory in the code repo in the course of testing (as is the current default behavior of `config.yaml`), it would get baked into the image. 
   - Perhaps I/we should add a .dockerignore file that ignores the `./data` folder.

docker-compose.yml
NortySpock: Makes sense. Do you want that `docker run` command example in this repos's README, or should I just write up a blurb you can add to the documentation (Or I could dig through the website repo to find where you have installation instructions and issue a PR there :) )

NortySpock:
Added .dockerignore file (same as the .gitignore file), and added a section in the readme on how to build and run a container and persist the data to a directory.

poundifdef:
Looks good! I'll give it a test and merge when I confirm. Appreciate your iterating on this :)

PR #165, opened 2024-04-17
Query as API

### LINKED ISSUE #164: Queries as an API
We want to launch a feature where users can create RESTful endpoints based on queries. This is an enhancement of "saved queries".

I have started this work in a branch: https://github.com/scratchdata/scratchdata/pull/163/files

This is roughly how I propose building it:

- Rename "Shared Query" to "Saved Query"
    - These queries can either be public, meaning they can be used as public share links
    - Or they can be private, meaning they require an API key to query
- Queries should accept named parameters. For example: `SELECT * FROM events WHERE user = @user`
    - These will be passed as query parms: `GET /user_events?user=alice`
- We should be able to create API keys which only have access to these queries (and not access to any other endpoints)
- When we create an API key, we should be able to set default parameters that cannot be overridden
    - For example, create an API key `api_key_1` with the `user` parameter set to `alice`.
    - In this case, the person making the query does not need to specify a `?user=alice`. If they do, then the parameter will be overwritten by the default
- Ideally, an API key could be granted access to multiple endpoints.

Finally, there should be a UI for creating these queries. 

- I had started a UI [prototype](https://github.com/scratchdata/api_builder/blob/main/api-builder/src/routes/create/%2Bpage.svelte): 
- Here's a [video](https://www.youtube.com/watch?v=XY5ytJMJGmc) of the ideal state.
- Homepage [screenshot](https://scratchdata.com/static/endpoints_screenshot.png).

If we have time:
- It would be awesome to be able to create models (protobuf or openapi) 
    - We can either use EXPLAIN statements or `DESCRIBE (select * from events where user = @user)` to get the output type. 
- Auto-generate documentation (just the basics)

breadchris:
Resolves #164

### DIFF
--- a/pkg/storage/database/models/models.go
+++ b/pkg/storage/database/models/models.go
@@ -8,13 +8,15 @@ import (
 	"gorm.io/gorm"
 )
 
-type ShareQuery struct {
+type SavedQuery struct {
 	gorm.Model
-	UUID          string `gorm:"index:idx_share_query_uuid,unique"`
+	UUID          string `gorm:"index:idx_saved_query_uuid,unique"`
 	DestinationID int64
 	Name          string
 	Query         string
 	ExpiresAt     time.Time
+	IsPublic      bool
+	Slug          string
 }
 
 type Team struct {
@@ -62,9 +64,14 @@ type ConnectionRequest struct {
 
 type APIKey struct {
 	gorm.Model
+	Name          string
 	DestinationID uint
 	Destination   Destination `gorm:"constraint:OnDelete:CASCADE"`
 	HashedAPIKey  string      `gorm:"index"`
+	// If the API key is specific to a saved query
+	SavedQueryID uint
+	SavedQuery   SavedQuery
+	QueryParams  datatypes.JSONMap
 }
 
 type MessageType string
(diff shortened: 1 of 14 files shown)


pkg/storage/database/models/models.go
```diff
@@ -62,9 +64,14 @@ type ConnectionRequest struct {
 
 type APIKey struct {
 	gorm.Model
+	Name          string
 	DestinationID uint
 	Destination   Destination `gorm:"constraint:OnDelete:CASCADE"`
 	HashedAPIKey  string      `gorm:"index"`
+	// If the API key is specific to a saved query
+	SavedQueryID uint
```
poundifdef, line 72: One comment here (I realize this is copied from my existing branch.) I'd started to refactor this so that the APIKey table does not have a SavedQuery, and instead use a Join table, ie, `apikey_savedquery`. Then I just added an APIKey "type" (`QUERY`, `GENERAL` - or something more descriptive) that tells me about who the key belongs to.

This also then lets me create a single API key and give it access to multiple endpoints. And it didn't make sense the `QueryParams` be in this table either since it's such a specific use case.

PR #173, opened 2024-05-09
Feature: Create new destination UI

CodeMan99:
/claim #170

Still testing. Overall pretty happy with this. ~~Video upload later tonight.~~

[Scratchdata Demo](http://www.youtube.com/watch?v=yc3RX6XM4yw "scratchdata demo")


frontend/src/lib/DestinationForm.svelte
```diff
@@ -0,0 +1,120 @@
...
+            headers: {
+                "content-type": "application/json",
+                "content-length": body.length,
+            },
+            body
+        };
+
+        const response = await fetch(resource, options);
+
+        if (response.ok) {
```
poundifdef, line 55: The code throws an error upon success. 

Screenshot 2024-05-09 at 7 09 14 PM

frontend/src/routes/dashboard/connections/new/[type]/+page.svelte
```diff
@@ -1 +0,0 @@
-<h2>New Connection</h2>
```
Comment (poundifdef) [inline frontend/src/routes/dashboard/connections/new/[type]/+page.svelte:1]: Please stick with the original implementation that uses this route instead of tabs for destination types. (I have a UI design that depends on these being different pages, which is a later iteration.)

frontend/src/lib/AlertAppError.svelte
```diff
@@ -0,0 +1,45 @@
...
+    export let duration = Infinity;
+    export let iconName = "exclamation-octagon";
+    export let ready = false;
+
+    export function show(error) {
+        errorMessage = error;
+        alertElement.show();
+    }
+
+    export function toast(error) {
```
poundifdef, line 20: This function doesn't seem to be used.

frontend/src/lib/AlertAppError.svelte
```diff
@@ -0,0 +1,45 @@
...
+        errorMessage = error;
+        alertElement.toast();
+    }
+
+    export function hide() {
```
poundifdef, line 25: This function doesn't seem to be used either.

poundifdef (changes requested):

frontend/src/lib/DestinationForm.svelte
```diff
@@ -0,0 +1,120 @@
...
+            {:else if formField.type === "textarea"}
+                <DestinationTextareaParam name="{formField.name}" label="{formField.label}" defaultValue="{formField.default}" />
+            {/if}
+        {/each}
+        <br>
+        <sl-button type="submit" variant="primary" outline pill>
+            <sl-icon name="cloud-plus" slot="prefix"></sl-icon>
+            Save New Connection
+        </sl-button>
+        <sl-button type="reset" variant="warning" outline pill>
```
poundifdef, line 110: I don't think we need a reset button here.

frontend/src/routes/dashboard/connections/new/+page.svelte
```diff
@@ -1,2 +1,73 @@
...
+                },
+            };
+            onTabShow(event);
+        }
+    });
+</script>
+
+<h2>New Connection</h2>
+
+<AlertAppError bind:this="{alertAppError}" />
```
poundifdef, line 45: Can you help me understand what ` bind:this="{alertAppError}"` is doing in this file? I'm not totally understanding how the `AlertAppError.svelte` component is supposed to work.

frontend/src/lib/AlertAppError.svelte
CodeMan99: This component is wrapping `<sl-alert>` for generic usage.

frontend/src/routes/dashboard/connections/new/+page.svelte
CodeMan99: This binds the entire component to the name of the parent component. It is later then passed into the component of each form.

Since you would like to have separate routes for each _connection type_, I will likely use a slightly different pattern anyway.

frontend/src/lib/DestinationForm.svelte
CodeMan99: A reset button is rarely necessary for typical users. It's more an accessibility utility.

frontend/src/lib/DestinationForm.svelte
CodeMan99: I've not used `redirect` before. I'll dig deeper.

frontend/src/routes/dashboard/connections/new/[type]/+page.svelte
Comment (CodeMan99) [reply frontend/src/routes/dashboard/connections/new/[type]/+page.svelte:1]: Ah, ok. Well it was fun to implement the lazy loading. I'll get on this as soon as possible.

Comment (poundifdef) [reply frontend/src/routes/dashboard/connections/new/[type]/+page.svelte:1]: For context, this is what the final result will look like. The goal is to get the basic functionality working with minimal styling, and then go in and do the pure design work.

ui

Comment (CodeMan99) [reply frontend/src/routes/dashboard/connections/new/[type]/+page.svelte:1]: I have very quickly restored the `[type]` route parameter. Definitely needs a bit more polish.

Before I continue, I want to be sure this is the desired behavior. What I am seeing is a separation of the menu and the form? That means an incorrect click will need to circle back around to the connections page in order to re-open the menu. It also means that the route can be tampered with by the user.

I'm all for implementing what you want. I just want to be sure that's what you _actually_ want?

frontend/src/lib/DestinationForm.svelte
CodeMan99: Replaced `redirect` call with `goto` utility. Seems solid.

frontend/src/routes/dashboard/connections/new/[type]/+page.svelte
Comment (CodeMan99) [reply frontend/src/routes/dashboard/connections/new/[type]/+page.svelte:1]: Fair enough.

poundifdef (approved):

